PR #5, opened 2024-12-02
Add CI - run tests using github actions.

treeform:


configs/a20_40x40.yaml
```diff
@@ -2,6 +2,7 @@
 
 defaults:
   - mettagrid
+  - _self_
```
daveey: what does this do?

requirements.txt
```diff
@@ -0,0 +1,19 @@
+hydra-core>=1.3.2
```
daveey: should these go into setup.py?

tests/test_metagrid.py
```diff
@@ -0,0 +1,110 @@
...
+
+import hydra
+import numpy as np
+from tqdm import tqdm
+import pandas as pd
+
+global actions
+global env
+
+def test_performance(env, actions, duration):
```
daveey: let's put these in different files. test_performance.py, test_determinism.py

tests/test_metagrid.py
```diff
@@ -0,0 +1,110 @@
...
+import os
+
+import time
+
+import hydra
+import numpy as np
+from tqdm import tqdm
+import pandas as pd
+
+global actions
```
daveey: do these need to be global? can we move these into test_performance?

tests/test_metagrid.py
```diff
@@ -0,0 +1,110 @@
...
+            tick += 1
+            if tick % 100 == 0:
+                pbar.update(time.time() - start - pbar.n)
+
+    print_stats(env._c_env.get_episode_stats())
+    print(f'SPS: {atns.shape[0] * tick / (time.time() - start):.2f}')
+
+actions = {}
+env = {}
+@hydra.main(version_base=None, config_path="../configs", config_name="a20_40x40")
```
daveey: i see you copied code from the perf test. that was a janky file i was manually using to test performance. for a unittest, we should simplify. we don't need to print anything. we should just run a perf test, and assert that the speed is reasonable

daveey (changes requested):

configs/a20_40x40.yaml
treeform: it does nothing, but it fixes a hydra warning, for some reason hydra wants this now.

tests/test_metagrid.py
treeform: yes I was just using this file as a wip

tests/test_metagrid.py
treeform: thats correct I just wanted a file that had some thing running.

my plan is to write this file to be more a test thing

I mostly worked on the build.yml

requirements.txt
treeform: I had issues with hydra and other libs. I will move them to setup.py eventually.

treeform: Turns out I need this to make "pufferlib" and "puffergrid" work. Once I fix setup.py in "pufferlib" and "puffergrid" requirements.txt could be removed but one thing at a time. I want to get this working.

configs/a20_40x40.yaml
treeform: > Explicitness: The _self_ keyword makes it explicit which part of the configuration is being overridden or extended. This reduces ambiguity, especially in large and complex configurations.

setup.py
```diff
@@ -49,22 +49,28 @@ def build_ext(srcs, module_name=None):
     packages=find_packages(),
     nthreads=num_threads,
     install_requires=[
-        "hydra-core",
-        "jmespath",
-        "matplotlib",
-        # "numpy==2.0.0",
-        "pettingzoo",
-	    "pufferlib",
+        "hydra-core>=1.3.2",
```
daveey: can you just read this from requirements.txt so we don't need 2 copies or places to edit?

tests/test_basic.py
```diff
@@ -0,0 +1,111 @@
...
+    print("mettaGridEnv.observation_space: ", mettaGridEnv.observation_space)
+    assert mettaGridEnv.observation_space.shape == (24, 11, 11)
+    print("mettaGridEnv.action_space: ", mettaGridEnv.action_space)
+    assert mettaGridEnv.action_space.nvec.tolist() == [8, 9]
+    print("mettaGridEnv.action_names(): ", mettaGridEnv.action_names())
+    assert mettaGridEnv.action_names() == [
+      'noop', 'move', 'rotate', 'use', 'attack', 'shield', 'gift', 'swap'
+    ]
+    print("mettaGridEnv.grid_features: ", mettaGridEnv.grid_features)
+    assert mettaGridEnv.grid_features == [
```
daveey: every time we add a property, we're gonna have to edit this file. let's not make people do that

tests/test_basic.py
```diff
@@ -0,0 +1,111 @@
...
+      'agent', 'agent:hp', 'agent:frozen', 'agent:energy', 'agent:orientation',
+      'agent:shield', 'agent:inv:r1', 'agent:inv:r2', 'agent:inv:r3', 'wall',
+      'wall:hp', 'generator', 'generator:hp', 'generator:r1', 'generator:ready',
+      'converter', 'converter:hp', 'converter:input_resource',
+      'converter:output_resource', 'converter:output_energy', 'converter:ready',
+      'altar', 'altar:hp', 'altar:ready'
+    ]
+    print("mettaGridEnv.global_features: ", mettaGridEnv.global_features)
```
daveey, line 106: i appreciate trying to have unittests, but this seems like it's just making work for people without providing much value.

daveey (changes requested):

setup.py
treeform: The requirements.txt will go a way once "pufferlib" and "puffergrid" setup.py works. Do you want me to switch to working on those repos instead? I did not want to overcomplicate this PR, but I can do that instead.

tests/test_basic.py
treeform: I could remove this if you like. But I appears that actions come from test_basic.yaml. If actions get added that are not part of test_basic.yaml that should fail this test?

tests/test_basic.py
treeform: I need an annoying unit test so that I make big changes to metagrid and puffergrid and make sure I break nothing. But I could see having two sets of tests one for normal use and one for big refactors.

requirements.txt
treeform: I changed my mind on this. I think requirements.txt is needed because `python setup.py build_ext --inplace` does not install any of the require libs only `python setup.py install` does. Also to run setup.py you need to have some dependancies already in place. I think its cleaner to just move all deps to requirements.txt. There are no extra copies of the deps.

setup.py
treeform: I removed the two copies. There is only one copy of all requirements.

tests/test_basic.py
treeform: I removed this assert for now.

tests/test_basic.py
treeform: I removed this assert for now.

daveey (approved):

PR #9, opened 2024-12-16
[WIP] Last action tracker

treeform:


mettagrid/wrappers/last_action_tracker.py
```diff
@@ -0,0 +1,61 @@
+import gymnasium as gym
```
daveey, line 1: this looks correct, but we discussed implementing this in cython for performance. Otherwise the step() function slows the env down significantly

mettagrid/wrappers/last_action_tracker.py
```diff
@@ -0,0 +1,61 @@
...
+        ]
+
+    @property
+    def observation_space(self):
+        box = self._env.observation_space
+        return Box(
+            0,
+            255,
+            (box.shape[0] + 2, box.shape[1], box.shape[2]),
+            np.uint8
```
daveey, line 60: grab the dtype from the existing shape, so that if we change it we don't have to update this code

mettagrid/wrappers/last_action_tracker.py
```diff
@@ -0,0 +1,61 @@
...
+        shape = obs.shape
+        last_action_feature = np.zeros(
+            (shape[0], 2, shape[2], shape[3]),
+            dtype=obs.dtype
+        )
+        # Set the middle of the last action features to the last action and
+        # last action argument:
+        for agentId in range(shape[0]):
+            last_action_feature[
+                agentId, :, shape[2]//2, shape[3]//2
```
daveey, line 40: avoid doing the // in the loop. move it to the constructor since the index stays the same.

daveey (changes requested):

mettagrid/wrappers/last_action_tracker.py
treeform: good idea!

mettagrid/wrappers/last_action_tracker.py
treeform: Yes. Doing it in python first is simpler. I want to get it working. There were so many things to figure out. Now that I have a python version working I can make the Cython version.

I do like your idea of "Feel free to edit grid_env.pyx directly if you think that's cleaner than adding a wrapper, just make sure we can turn it off via config". I'll try that too.

mettagrid/wrappers/last_action_tracker.py
treeform: I'll try that.

I lose a ton of speed resizing and creating the new tensors. I feel like I need another hook during their creation time so that they can be made the right size from the start.

mettagrid/wrappers/last_action_tracker.py
treeform: maybe single_observation_space?

PR #20, opened 2025-01-06
Allow for ascii maps

daphnedemekas:
Change game_builder so we can parse ascii maps


daphnedemekas:
Currently failing the test because this requires a new default in the mettagrid.yaml file (can set it to map: file: false)

mettagrid/mettagrid_env.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, render_mode: str, buf=None, **cfg):
         self.should_reset = False
         self._renderer = None
 
+
```
daveey: extra white space

mettagrid/config/game_builder.py
```diff
@@ -46,11 +49,72 @@ def __init__(
...
+                elif char == "g":
+                    level[i,j] = self._symbols["generator"]
+                    generator_count += 1
+                elif char == " ":
+                    level[i,j] = self._symbols["empty"]
+
+        self.num_agents = agent_count
+        self.obs_width = width
+        self.obs_height = height
+#
```
daveey: what's this?

mettagrid/config/game_builder.py
```diff
@@ -46,11 +49,72 @@ def __init__(
     def level(self):
         layout = self.map_config.layout
 
-        if "rooms" in layout:
+        if self.map_config.file:
```
daveey: can you add this to mettagrid.yaml as well? that way this will pass tests

mettagrid/config/game_builder.py
```diff
@@ -19,6 +19,9 @@ def __init__(
             objects,
             actions,
             map):
+        
+        #map.file -> ascii map
```
daveey: what's this?

mettagrid/config/game_builder.py
```diff
@@ -46,11 +49,72 @@ def __init__(
...
+
+        # Track number of agents found
+        agent_count = 0
+        wall_count = 0
+        altar_count = 0
+        converter_count = 0
+        generator_count = 0
+
+        # Parse each character in the ASCII map
+        for i, line in enumerate(lines):
```
daveey: Are you just translating a hardcoded symbol to the same symbol? Is this whole thing needed? why can't you just convert the lines to a numpy array and return that?

daveey (changes requested):

mettagrid/config/game_builder.py
daphnedemekas: sorry I thought at first that I would have to update these arguments based on info from the map (also why the code below was with a bunch of if statements) but i realize that's not necessary

mettagrid/config/game_builder.py
daphnedemekas: Same comment as above, I thought we needed to keep track of counts but we don't! Changed now to single line

PR #21, opened 2025-01-07
Add last action trackering.

treeform:


configs/mettagrid.yaml
```diff
@@ -107,6 +107,8 @@ game:
         generator: 15
         wall: 10
 
+last_action_tracker: false
```
daveey: track_last_action

tests/bench_env.py
```diff
@@ -0,0 +1,115 @@
+import time
```
daveey: test_env_perf or benchmark_env_perf

daveey (changes requested):

tests/bench_env.py
daveey: regen_env_trace.py

daveey: gold/env_trace.txt

daveey (approved):

PR #22, opened 2025-01-07
Add .close() to mettagrid so it does not crash.

treeform:
When using:

```sh
python -m tools.train run=my_experiment hardware=macbook trainer.total_timesteps=2
```

 this prevents crashing on close because close is not implemented.


daveey (commented):
lgtm

PR #30, opened 2025-01-17
Track action_success.

treeform:
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209257698169585) by [Unito](https://www.unito.io)

### DIFF
--- a/mettagrid/actions/actions.pyx
+++ b/mettagrid/actions/actions.pyx
@@ -51,11 +51,9 @@ cdef class MettaActionHandler(ActionHandler):
             strcat(stat_name, ".status.frozen.ticks")
             self.env._stats.agent_incr(actor_id, stat_name)
             actor.frozen -= 1
-            self.env._action_success[actor_id] = False
             return False
 
         if actor.energy < self.action_cost:
-            self.env._action_success[actor_id] = False
             return False
 
         actor.update_energy(-self.action_cost, &self.env._rewards[actor_id])
--- a/mettagrid/actions/actions.pyx
+++ b/mettagrid/actions/actions.pyx
@@ -70,7 +70,6 @@ cdef class MettaActionHandler(ActionHandler):
         if result:
             self.env._stats.agent_incr(actor_id, self._stats.action.c_str())
 
-        self.env._action_success[actor_id] = result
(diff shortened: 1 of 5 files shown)


mettagrid/actions/actions.pyx
```diff
@@ -68,6 +70,7 @@ cdef class MettaActionHandler(ActionHandler):
         if result:
             self.env._stats.agent_incr(actor_id, self._stats.action.c_str())
 
+        self.env._action_success[actor_id] = result
```
daveey: why are you doing this here, instead of in grid_env.pyx where we return the result?

mettagrid/grid_env.pyx
```diff
@@ -222,6 +222,7 @@ cdef class GridEnv:
         self._rewards = rewards
         self._episode_rewards_np = np.zeros_like(rewards)
         self._episode_rewards = self._episode_rewards_np
+        self._action_success = np.zeros(len(rewards), dtype=np.int8)
```
daveey: you can just use a vector[bint] here, no need for numpy

daveey: and set in the constructor rather than in set_buffers

daveey (changes requested):

mettagrid/actions/actions.pyx
treeform: Because all the other `self.env._action_success[actor_id]` happen in this file. It creates less cognitive load when everything related is closer together. 
But I can move it to a different file if you like?

treeform: > Can you update action_succcess in grid _env instead?

Done.

mettagrid/actions/actions.pyx
```diff
@@ -51,9 +51,11 @@ cdef class MettaActionHandler(ActionHandler):
             strcat(stat_name, ".status.frozen.ticks")
             self.env._stats.agent_incr(actor_id, stat_name)
             actor.frozen -= 1
+            self.env._action_success[actor_id] = False
```
daveey: remove

daveey (approved):

PR #33, opened 2025-01-25
Configurable converters

sasmith:
This adds configurable converters. I've been sitting on the code for a while, but figured I'd make a draft while I think through pieces.

Right now there's a single, hard coded configuration that replaces existing converters. Heart altars should be able to be replaces with the same object, different recipe. Generators can almost be replaced, with the exception that generators can accumulate things you can pick up, and the current implementation of recipes doesn't offer that.

I'm not sure what to do about switching recipes, of recipe visibility. It seems reasonable to see the recipe that a converter is set to, and maybe have a way of seeing what that recipe does. I know how to encode the former, though the numeric nature of the variable will be misleading. I'm not sure what to do about giving info about what recipes do.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209257697502252) by [Unito](https://www.unito.io)

### DIFF
--- a/mettagrid/actions/use.pyx
+++ b/mettagrid/actions/use.pyx
@@ -6,7 +6,7 @@ from omegaconf import OmegaConf
 from mettagrid.grid_object cimport GridLocation, GridObjectId, Orientation, GridObject
 from mettagrid.action cimport ActionHandler, ActionArg
 from mettagrid.objects cimport MettaObject, ObjectType, Usable, Altar, Agent, Events, GridLayer
-from mettagrid.objects cimport Generator, Converter, InventoryItem, ObjectTypeNames, InventoryItemNames
+from mettagrid.objects cimport Generator, Converter, ConverterRecipe, InventoryItem, ObjectTypeNames, InventoryItemNames
 from mettagrid.actions.actions cimport MettaActionHandler
 
 cdef class Use(MettaActionHandler):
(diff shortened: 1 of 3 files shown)


mettagrid/actions/use.pyx
```diff
@@ -55,17 +55,19 @@ cdef class Use(MettaActionHandler):
             self.env._stats.game_incr("r1.harvested")
 
         cdef Converter *converter
+        cdef ConverterRecipe recipe
```
daveey, line 58: unless you want to copy the recipe every time, use a ptr

mettagrid/actions/use.pyx
```diff
@@ -55,17 +55,19 @@ cdef class Use(MettaActionHandler):
...
-            actor.update_inventory(converter.output_resource, 1)
-            self.env._stats.agent_incr(actor_id, InventoryItemNames[converter.output_resource] + ".gained")
-
-            energy_gain = actor.update_energy(converter.output_energy, &self.env._rewards[actor_id])
-
-            self.env._stats.agent_add(actor_id, "energy.gained", energy_gain)
+            recipe = converter.recipes[0]
+            actor.update_energy(recipe.delta_energy, &self.env._rewards[actor_id])
+            for i in range(InventoryItem.InventoryCount):
+                if recipe.delta_resources[i] > 0:
```
daveey, line 65: you could pass the converter to the Recipe constructor. Then here call recipe.use(agent). That way the logic can live in the recipe rather than in the "use" action

PR #34, opened 2025-01-27
Some renaming and make_odd

daphnedemekas:
Note: planning on changing the maze algos anyway so this is not super important

For mazes, the requirement for odd dimensions comes from the fact that the algorithm creates corridors by moving in steps of 2 cells with walls in between. This ensures we always have proper walls between corridors and no adjacent corridors.

I changed it so that if we pass in even numbers it just converts them odd from within the maze class rather than in the yaml file. If we did a padding after the maze was constructed it would obstruct some of the maze. 

Do we want to be able to have even mazes? It would require some changes to the algorithm, which happy to make but just wondering if that's necessary

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209257796290569) by [Unito](https://www.unito.io)

### DIFF
--- a/mettagrid/config/room/maze.py
+++ b/mettagrid/config/room/maze.py
@@ -4,6 +4,10 @@ import numpy as np
 
 from mettagrid.config.room.room import OBJECTS, Room
 
+def make_odd(x):
+    if x % 2 == 0:
+        x += 1
+
 class Maze(Room):
     EMPTY, WALL = OBJECTS.Empty.symbol, OBJECTS.Wall.symbol
     START, END = OBJECTS.Agent.symbol, OBJECTS.Altar.symbol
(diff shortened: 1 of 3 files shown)


mettagrid/config/room/maze.py
```diff
@@ -4,6 +4,10 @@
 
 from mettagrid.config.room.room import OBJECTS, Room
 
+def make_odd(x):
+    if x % 2 == 0:
+        x += 1
```
daveey, line 9: i don't think we should make the room larger. since the container is telling it how big to be. we can make it smaller, and fill the missing row/column with walls. this means width=10 and width=9 would make the same maze, but with an extra wall column.

daveey (changes requested):

mettagrid/config/room/maze.py
daveey: once you support even dims, can you also remove the asserts and the use of make_odd() in the configs?

daphnedemekas: I think I'm understanding you, in which case if I did this, the extra wall column would obstruct the structure of the maze

daveey: also merge main

PR #37, opened 2025-02-04
Fix maze dimensionality issue

daphnedemekas:
This is a replacement for this PR https://github.com/Metta-AI/mettagrid/pull/34 
which had issues due to new updates

just a simple fix to maintain odd dimensionality without increasing the grid size but rather padding the grid

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209314652754677) by [Unito](https://www.unito.io)

### DIFF
--- a/configs/maze.yaml
+++ b/configs/maze.yaml
@@ -21,9 +21,9 @@ game:
     border_width: 2
     room:
       _target_: mettagrid.config.room.maze.Maze
-      width: ${uniform:10,40,21}
-      height: ${uniform:10,40,21}
-      start_pos: [1, 1]
+      width: ${uniform:10,40,20}
+      height: ${uniform:10,40,20}
+      start_pos: [1, 0]
       end_pos:
         - ${sub:${..width},2}
         - ${sub:${..height},2}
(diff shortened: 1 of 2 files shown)


daveey:
I think this can be much simpler. 

in init():

if width % 2, width=width-1, pad_column=True
same for height

generate maze as previously, now with odd dimension
then add column and row padding as needed

daphnedemekas:
What is it that you don't like specifically? In terms of padding, I am just making the effective width and height 1 less when I have even dimensions, and the padding happens in the final maze

```
        final_maze = np.full((self._height, self._width), self.WALL, dtype='<U50')
        # Copy the effective maze into the top-left corner of the container.
        final_maze[:self.eff_height, :self.eff_width] = maze_eff
```
        
which seems pretty simple but if you'd rather the padding be different I can change it. 

If it's the functions at the top that you don't like (the make odd and set position), those are with respect to the start and end positions, which need to be odd in order to be in the maze at all, and within the maze bounds. 

If I am just changing the effective width of the maze by subtracting by 1, and the end position was inputted as the bottom right corner, this would result in the end position being on the padded wall, if we padded the right side. Same with start position, if I put in (0,0) and padded the left side, then my start position would be in the wall. 

So that's why I am adding in functionality to make sure that these start and end positions are odd and within the effective maze bounds, to prevent issues like that

daveey (approved):

PR #39, opened 2025-02-05
Add an attack nearest action

sasmith:
This action attacks the nearest agent within the normal 3x3 attack grid. The intention here is to make attacking easier to learn. It currently uses the same config as attacks.

The main thing I want to test is how much this is used vs attack. Presumably the improved accuracy of this vs attack should mean it's favored by policies, and replaces attacks.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1209230969804461/1209335421215149) by [Unito](https://www.unito.io)

### DIFF
--- a/mettagrid/actions/attack_nearest.pyx
+++ b/mettagrid/actions/attack_nearest.pyx
@@ -13,7 +13,7 @@ from mettagrid.actions.attack cimport Attack
 
 cdef class AttackNearest(Attack):
     def __init__(self, cfg: OmegaConf):
-        MettaActionHandler.__init__(self, cfg, "attack_nearest")
+        Attack.__init__(self, cfg, "attack_nearest")
         self._priority = 1
 
     cdef unsigned char max_arg(self):
--- a/mettagrid/actions/attack_nearest.pyx
+++ b/mettagrid/actions/attack_nearest.pyx
@@ -1,0 +1,47 @@
+
+from libc.stdio cimport printf
+from libc.string cimport strcat, strcpy
+
+from omegaconf import OmegaConf
+
+from mettagrid.grid_object cimport GridLocation, GridObjectId, Orientation, GridObject
+from mettagrid.action cimport ActionHandler, ActionArg
+from mettagrid.objects cimport MettaObject, ObjectType, Usable, Altar, Agent, Events, GridLayer
+from mettagrid.objects cimport Generator, Converter, InventoryItem, ObjectTypeNames, InventoryItemNames
+from mettagrid.actions.actions cimport MettaActionHandler
+from mettagrid.actions.attack cimport Attack
+
+cdef class AttackNearest(Attack):
+    def __init__(self, cfg: OmegaConf):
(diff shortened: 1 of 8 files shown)


mettagrid/actions/attack_nearest.pyx
```diff
@@ -0,0 +1,47 @@
...
+        Agent * actor,
+        ActionArg arg):
+
+        cdef int distance = 0
+        cdef int offset = 0
+        cdef GridLocation target_loc;
+        cdef Agent * agent_target;
+        
+        # xcxc see if this is efficient
+        for distance in range(1, 4):
```
sasmith, line 33: I'll want to double check this and the next part of the loop don't do anything funny in Python.

mettagrid/actions/attack_nearest.pyx
```diff
@@ -0,0 +1,47 @@
...
+from mettagrid.grid_object cimport GridLocation, GridObjectId, Orientation, GridObject
+from mettagrid.action cimport ActionHandler, ActionArg
+from mettagrid.objects cimport MettaObject, ObjectType, Usable, Altar, Agent, Events, GridLayer
+from mettagrid.objects cimport Generator, Converter, InventoryItem, ObjectTypeNames, InventoryItemNames
+from mettagrid.actions.actions cimport MettaActionHandler
+from mettagrid.actions.attack cimport Attack
+
+cdef class AttackNearest(Attack):
+    def __init__(self, cfg: OmegaConf):
+        MettaActionHandler.__init__(self, cfg, "attack_nearest")
```
daveey: nbd but you should init Attack rather than MettaActionHadler

mettagrid/actions/attack_nearest.pyx
daveey: this should be fine, cdef distance is the key

mettagrid/actions/swap.pyx
```diff
@@ -26,6 +26,8 @@ cdef class Swap(MettaActionHandler):
             actor.location,
         <Orientation>actor.orientation
         )
+        # xcxc we should ensure this is an agent. We're not just looking on the agent layer.
+        # Maybe the assertion that the target is frozen is enough?
```
daveey: relative_location is relative to the actor's location, which is an agent. so it automatically looks at the agent layer.

daveey (approved):
looks great

mettagrid/actions/attack_nearest.pyx
```diff
@@ -0,0 +1,48 @@
...
+        ActionArg arg):
+
+        cdef int distance = 0
+        cdef int offset = 0
+        cdef GridLocation target_loc;
+        cdef Agent * agent_target;
+        
+        # Scan the space to find the nearest agent. Prefer the middle (offset 0) before the edges (offset -1, 1).
+        for distance in range(1, 4):
+            for offset in range(3):
```
sasmith, line 34: The previous version of this (`for distance in [0, -1, 1]`) went into Python land.

PR #42, opened 2025-02-12
New maps for cogevals

georgejwdeane:
This contains python files for new cognitive eval maps. 
Once this is merged in, we can train cognitive evals on AWS


daphnedemekas:
@georgejwdeane Can you please move all of the cognitive eval python files into `mettagrid/config/room/cognitive_evals` and then make sure that your config files in the metta repository correspond to the new paths 

Once that's done this should be ready for merge

daphnedemekas (changes requested):
Can you please move all of the cognitive eval python files into mettagrid/config/room/cognitive_evals and then make sure that your config files in the metta repository correspond to the new paths

Once that's done this should be ready for merge

PR #43, opened 2025-02-14
Split objects.pxd into separate files per object

sasmith:
This turned out to mostly be a battle with cython. I'm not confident that all of the lessons I learned are correct, since I could definitely be conflating issues. I had a lot of flailing, but two failures stand out

Using an __init__.pxd to make imports nicer.
I wanted to have people continue to be able to do `from mettagrid.objects import Foo` instead of `from mettagrid.objects.foo import Foo`, and I intended to do this by having `objects/__init__.pxd` that imported the relevant files. My intention was to also just build this single pxd, so there would be a single `objects` module that you imported (more or less).

This failed with errors around `PyInit____init__` not being defined for the module (runtime issue)

Using .pyx files.
When I had implementations in .pyx files, I ended up with a lot of runtime errors about missing symbols. This very much appears to be a linking issue. I only reliably had this show up when using cppclass, and subclassing, so I have a hazy theory: I think normally symbols are picked up because python more-or-less does it's normal importing, as a version of dynamic linking, but that with cpp class you end up needing to use the symbols too early in this process and crash before you can do the import.

I could solve this problem by having `foo.pyx` list `bar.pyx` as a source file, but then ran into issues with the symbols in the `pxd` files being doubly-declared (at compile time).

Takeaways:
1) Stick with header-only when working with cython, at least when working with cppclass.
2) Bolster the above by just moving stuff from cython -> cpp. Here I think we'll still need to deal with linking issues, but it'll be a lot more straightforward.


mettagrid/actions/use.pyx
```diff
@@ -33,13 +37,14 @@ cdef class Use(MettaActionHandler):
         )
         target_loc.layer = GridLayer.Object_Layer
         cdef MettaObject *target = <MettaObject*>self.env._grid.object_at(target_loc)
-        if target == NULL:
+        if target == NULL or not target.is_usable_type():
```
sasmith, line 40: This allows us to not need to define `usable` on `metta_object`, and thus make `metta_object` not depend on `Agent`.

sasmith: I now (without testing) assume this won't work. `is_usable_type` is inline (so it can be in the header -- a cython thing), which seems like it must make it non-virtual. So we're always going to get the base version of the function here (it's inline -- how could it be polymorphic?).

Not testing now, since I'm converting things to cpp which I think will solve this. I'll test post-expected-solve.

daveey (approved):

sasmith:
Merged https://github.com/Metta-AI/mettagrid/pull/45, which includes this.

PR #45, opened 2025-02-15
Split objects.pyx; move parts to cpp

sasmith:
This is intended to be a noop code cleanup. It's an extension of https://github.com/Metta-AI/mettagrid/pull/43 that also moves more logic into c++. The fact that I'm just using header files in c++ isn't intended as any long term statement -- just that this is the easiest translation from using pxds.

For posterity the commit message of https://github.com/Metta-AI/mettagrid/pull/43 is:

This turned out to mostly be a battle with cython. I'm not confident that all of the lessons I learned are correct, since I could definitely be conflating issues. I had a lot of flailing, but two failures stand out

Using an init.pxd to make imports nicer.
I wanted to have people continue to be able to do from mettagrid.objects import Foo instead of from mettagrid.objects.foo import Foo, and I intended to do this by having objects/__init__.pxd that imported the relevant files. My intention was to also just build this single pxd, so there would be a single objects module that you imported (more or less).

This failed with errors around PyInit____init__ not being defined for the module (runtime issue)

Using .pyx files.
When I had implementations in .pyx files, I ended up with a lot of runtime errors about missing symbols. This very much appears to be a linking issue. I only reliably had this show up when using cppclass, and subclassing, so I have a hazy theory: I think normally symbols are picked up because python more-or-less does it's normal importing, as a version of dynamic linking, but that with cpp class you end up needing to use the symbols too early in this process and crash before you can do the import.

I could solve this problem by having foo.pyx list bar.pyx as a source file, but then ran into issues with the symbols in the pxd files being doubly-declared (at compile time).

Takeaways:

Stick with header-only when working with cython, at least when working with cppclass.
Bolster the above by just moving stuff from cython -> cpp. Here I think we'll still need to deal with linking issues, but it'll be a lot more straightforward.


sasmith:
Merging as this is (intended as) a noop. I've tested and trained https://wandb.ai/metta-research/metta/runs/sasmith.teamroom.freeze.50?nw=nwusersasmithstemai as a smoke test.

PR #49, opened 2025-02-25
Clean up imports and things

sasmith:
This makes objects more consistent by cleaning up imports, build directives, inline-ness of functions, and some other stuff. I think it's mostly all in the right direction (I think we might want `except +` added to other functions too?), but in any case, this is mostly a push for consistency.


sasmith:
Merging as this should practically be a noop.

PR #50, opened 2025-02-25
clean up obs functions

sasmith:
We previously defined an inline observation function per object type, and then needed to use _type_id to figure out what casting to do in observation_encoder. I believe we had to do that since we'd been writing all functions in cython header files, and this requires the functions to be inline, and thus not virtual. Now that we're using (more) proper cpp, we can use virtual functions, and let someone else figure out which function we should be calling.

Tested this by playing mettagrid and mousing over things, and confirming that the observations still looked right.

### DIFF
--- a/mettagrid/grid_object.pxd
+++ b/mettagrid/grid_object.pxd
@@ -21,6 +21,8 @@ cdef extern from "grid_object.hpp":
 
     ctypedef unsigned int GridObjectId
 
+    ctypedef unsigned char ObsType
+
     cdef cppclass GridObject:
         GridObjectId id
         GridLocation location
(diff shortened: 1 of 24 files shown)


mettagrid/grid_object.pxd
```diff
@@ -21,6 +21,8 @@ cdef extern from "grid_object.hpp":
 
     ctypedef unsigned int GridObjectId
 
+    ctypedef unsigned char ObsType
```
sasmith, line 24: Moved this from base_encoder to remove a circular dependency.

sasmith:
Merging as this should be a noop.

PR #52, opened 2025-02-26
Adds SemiCompactObservations

sasmith:
This changes observation encoding to allow a vector of offsets, rather than a single offset, when encoding observations. This allows us to do non-sequential observation encoding. For both existing encoders, this just means a little more work, since they only use sequential observation space.

This also adds a SemiCompactObservationEncoder, which uses these new offsets to allow us to encode same-named observations to the same space. I.e., we can record "hp" and "ready" to the same location in observation space for all objects.

My un-tested assumption is that this is a more useful encoding for the agent. Moreover, it reduces our feature list from 38 to 22. This will also make it less expensive to give a visible inventory to all buildings, since it will add ~6 features names instead of adding ~50.

This compiles and trains, both with and without the semi_compact_obs flag. So far this is only a "it doesn't crash" smoke test.


mettagrid/grid_object.pxd
```diff
@@ -31,5 +32,5 @@ cdef extern from "grid_object.hpp":
         GridObject()
         void __dealloc__()
         void init(TypeId type_id, const GridLocation &loc)
-        void obs(ObsType *obs)
+        void obs(ObsType *obs, vector[unsigned int] offsets)
```
daveey: you probably want const vector[unisgned int] &offsets, otherwise you copy the vector on every call

mettagrid/objects/agent.hpp
```diff
@@ -93,24 +93,24 @@ class Agent : public MettaObject {
...
 
-    virtual void obs(ObsType* obs) const override {
-        obs[0] = 1;
-        obs[1] = group;
-        obs[2] = hp;
-        obs[3] = frozen;
-        obs[4] = orientation;
-        obs[5] = color;
+    virtual void obs(ObsType* obs, std::vector<unsigned int> offsets) const override {
+        obs[offsets[0]] = 1;
```
daveey, line 97: unsigned char o = 0;
obs[offsets[o++]] = 1
....

mettagrid/observation_encoder.pyx
```diff
@@ -18,57 +18,101 @@ from mettagrid.objects.generator cimport Generator
...
+            for i in range(len(self._type_feature_names[type_id])):
+                self._offsets[type_id][i] = features[self._type_feature_names[type_id][i]]
+
+        self._feature_names.resize(features.size())
+        cdef map[string, int].iterator it = features.begin()
+        for i in range(features.size()):
+            self._feature_names[i] = dereference(it).first
+            postincrement(it)
+
+cdef class CompactObservationEncoder(ObservationEncoder):
```
daveey: we don't use compact, and i dont think it works, just kill it

daveey (changes requested):
feel free to merge after fixing the offsets to be a reference

sasmith:
When I tested using `ObservationEncoder` (non-compact), I initially seemed to see worse results (_much_ lower average reward after 1-3 gigasteps).

This didn't clearly repro when I tried to debug. Observations look correct, and at best the bug seems to correlate with me changing feature names.

The only training-related thing that feature names should impact is normalization. Testing "just changing normalization" didn't obviously show issues.

Merging based on "no statistically significant demonstration of a bug"; but I'm still a bit wary.

PR #53, opened 2025-02-27
Add converter inventories / queues

sasmith:
Makes all buildings be Converters, and gives them an inventory. Converters have a recipe, and based on this they'll convert their inventory from one type to another.

Right now most conversion parameters are hard coded.

My main testing was with much shorter cooldowns, but it felt satisfying to be able to pick up and drop off a load of ore.

Follow ups:
* Make more of the conversion configurable.
* Clean up how conversion is triggered, and how inventory is accessed.
* Make put / get work on agents.
* Ensure that tracing works with the new actions.


configs/mettagrid.yaml
```diff
@@ -129,6 +129,12 @@ game:
     transfer:
       enabled: false
       cost: 0
+    put_action:
```
daveey, line 132: i don't love the name, how about "put_items" / "get_items"

daveey: we no-longer have action costs, so we could turn this into 

disabled_actions = []

mettagrid/actions/get_output.pyx
```diff
@@ -0,0 +1,58 @@
...
+    cdef unsigned char max_arg(self):
+        return 0
+
+    cdef bint _handle_action(
+        self,
+        unsigned int actor_id,
+        Agent * actor,
+        ActionArg arg):
+
+        cdef GridLocation target_loc = self.env._grid.relative_location(
```
daveey, line 25: we do this a lot, might be worth having self._target() defined on MettaActionHandler

mettagrid/actions/put_recipe.pyx
```diff
@@ -0,0 +1,50 @@
...
+from mettagrid.grid_object cimport GridLocation, Orientation
+from mettagrid.action cimport ActionArg
+from mettagrid.objects.agent cimport Agent
+from mettagrid.objects.metta_object cimport MettaObject
+from mettagrid.objects.constants cimport Events, GridLayer, InventoryItem, InventoryItemNames
+from mettagrid.objects.converter cimport Converter
+from mettagrid.actions.actions cimport MettaActionHandler
+
+# Puts one recipe worth of resources into a Converter. Noop if not enough resources.
+cdef class PutRecipe(MettaActionHandler):
```
daveey: I read this as "insert a recipe into the converter", maybe PutRecipeItems ?

mettagrid/actions/put_recipe.pyx
```diff
@@ -0,0 +1,50 @@
...
+
+        # #Converter_and_HasInventory_are_the_same_thing
+        cdef Converter *converter = <Converter*> target
+
+        for i in range(converter.recipe_input.size()):
+            if converter.recipe_input[i] > actor.inventory[i]:
+                return False
+
+        for i in range(converter.recipe_input.size()):
+            converter.inventory[i] += converter.recipe_input[i]
```
daveey: do we not want to call converter.update_inventory()?

daveey (approved):

mettagrid/actions/put_recipe.pyx
sasmith: We do, but we want that function to be void, and we want to pass it whatever it needs in order to schedule production finishing. Since we can't do the latter [without more work], I didn't bother doing the former in this diff. I.e., there isn't an update_inventory function yet.

configs/mettagrid.yaml
sasmith: will clean up separately

mettagrid/actions/get_output.pyx
sasmith: Seems reasonable, but I think we only do this twice (swap is / should be somewhat different), so I'll leave this as clean up for when it's closer to something should should happen for all actions.

mettagrid/actions/put_recipe.pyx
sasmith: done

PR #55, opened 2025-03-01
Convert events to be cpp

sasmith:
Generally we want to move things out of cython, so that's the main motivation here. The work was getting things to not depend on env, and then to untangle the "when can things be Python objects vs cpp objects vs pointers", which took some stumbling on my part.


mettagrid/event.pxd
```diff
@@ -12,27 +13,28 @@ cdef extern from "event.hpp":
...
+    cdef cppclass EventHandler:
+        EventManager *event_manager
 
-cdef class EventHandler:
-    cdef GridEnv env
-    cdef EventId event_id
+        EventHandler(EventManager *event_manager)
 
-    cdef void init(self, GridEnv env, EventId event_id)
-    cdef void schedule(self, unsigned int delay, GridObjectId object_id, EventArg arg)
```
sasmith, line 37: dropped this functionality since it's unused

daveey (approved):

PR #58, opened 2025-03-18
add recipe_duration as distinct from cooldown, and initial_items

daveey:
Converters now can start with initial_items worth of output. They also take conversion_ticks to convert, and then wait cooldown to start converting again. This allows control of supply separately from credit assignment delay

### DIFF
--- a/mettagrid/mettagrid.pyx
+++ b/mettagrid/mettagrid.pyx
@@ -25,6 +25,7 @@ from mettagrid.objects.mine cimport Mine
 from mettagrid.objects.agent cimport Agent
 from mettagrid.objects.production_handler cimport ProductionHandler
 from mettagrid.objects.wall cimport Wall
+from mettagrid.objects.converter cimport Converter
 from mettagrid.objects.generator cimport Generator
 from mettagrid.objects.altar cimport Altar
 from mettagrid.objects.lab cimport Lab
@@ -32,7 +33,8 @@ from mettagrid.objects.factory cimport Factory
 from mettagrid.objects.temple cimport Temple
 from mettagrid.objects.armory cimport Armory
 from mettagrid.objects.lasery cimport Lasery
-from mettagrid.objects.constants cimport ObjectLayers, InventoryItemNames, Events
+from mettagrid.objects.constants cimport ObjectLayers, InventoryItemNames
+
 
 # Action imports
 from mettagrid.actions.move import Move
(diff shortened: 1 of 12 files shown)


mettagrid/mettagrid.pyx
```diff
@@ -175,6 +148,16 @@ cdef class MettaGrid(GridEnv):
                     agent.agent_id = self._agents.size()
                     self.add_agent(agent)
                     self._group_sizes[group_id] += 1
+
+                if converter != NULL:
+                    stat = "objects." + map[r,c]
```
sasmith, line 153: next time we refactor, it would be nice to pull out this logic to cover walls and agents too.

mettagrid/objects/converter.hpp
```diff
@@ -73,8 +75,36 @@ class Converter : public HasInventory {
...
+            for (unsigned int i = 0; i < InventoryItem::InventoryCount; i++) {
+                if (this->recipe_output[i] > 0) {
+                    HasInventory::update_inventory(static_cast<InventoryItem>(i), initial_items, nullptr);
+                }
+            }
+        } else {
+            // Use recipe_output amounts as initial values if initial_items not specified
+            for (unsigned int i = 0; i < InventoryItem::InventoryCount; i++) {
+                if (this->recipe_output[i] > 0) {
+                    HasInventory::update_inventory(static_cast<InventoryItem>(i), this->recipe_output[i], nullptr);
```
sasmith, line 104: I don't understand the impact of changing from `this->update_inventory` to `HasInventory::update_inventory`. Can you give me a sentence or two?

sasmith (approved):
> This allows control of supply separately from credit assignment delay

I don't understand this.

Starting inventory makes sense. I don't really understand the need for cooldowns separately from production times. The main thing I see this allowing is for you to set a production time of zero (or 1), and use cooldowns to have this not be way too fast on net. And this would be to improve the breadcrumbs for the agents. Is that right? Is there some other use case?

mettagrid/mettagrid.pyx
daveey: agreed

mettagrid/objects/converter.hpp
daveey: update_inventory also calls maybe_start_converting, which we don't want to do in this case. that said, i made sure cfg contains all the necessary fields, so i'll simplify the logic to not check for misgging values

daveey (commented):
the main reason for recipe_duration is that the delay between an agent taking an action, and the agent seeing some useful change to the environment effects their ability to learn. this decouples the production rate, from the time to see impact

PR #59, opened 2025-03-19
Colored ore

daphnedemekas:
Colored ore, and swappable blocks


daveey (approved):

mettagrid/objects/constants.hpp
```diff
@@ -55,8 +59,13 @@ std::vector<std::string> InventoryItemNames = {
 std::vector<std::string> ObjectTypeNames = {
     "agent",
     "wall",
-    "mine",
-    "generator",
+    "block",
```
sasmith, line 62: Hello from the future! This smells a little funny to me, since this vector no longer aligns with ObjectType. It's causing some issues for me, but I think it just generically breaks some naming. I expect to address this, but lmk if you think I'm missing something.

mettagrid/mettagrid.pyx
```diff
@@ -114,11 +114,20 @@ cdef class MettaGrid(GridEnv):
...
                     self._grid.add_object(wall)
                     self._stats.incr(b"objects.wall")
-
-                elif map[r,c] == "mine":
-                    converter = new Mine(r, c, cfg.objects.mine)
-                elif map[r,c] == "generator":
-                    converter = new Generator(r, c, cfg.objects.generator)
+                elif map[r,c] == "block":
+                    block = new Wall(r, c, cfg.objects.block)
+                    self._grid.add_object(block)
```
daphnedemekas, line 119: @sasmith a block is a wall type, with swappable enabled

mettagrid/objects/constants.hpp
daphnedemekas: hey, see my comment in mettagrid.pyx. Are you talking about the addition of block here? That is the object type of Wall, with the swappable enabled

PR #71, opened 2025-04-07
Training suite

daphnedemekas:


mettagrid/mettagrid_env.py
```diff
@@ -168,11 +171,49 @@ def object_type_names(self):
...
+    if config_path.startswith("/"):
+        config_path = config_path[1:]
+    path = config_path.split("/")
+    for p in path[:-1]:
+        env_cfg = env_cfg[p]
+    if overrides is not None:
+        try:
+            env_cfg = OmegaConf.merge(env_cfg, overrides)
+        except ConfigKeyError:
+            logger.warning(f"Cannot parse overrides when using multienv_mettagrid")
```
sasmith: If we fail to parse the config, is `env_cfg` in a shape where we should be returning it? I would have expected us to still want to throw in this case.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +171,49 @@ def object_type_names(self):
...
+        env_cfg = self.select_env()
+        super().make_env(env_cfg)
+
+
 def make_env_from_cfg(cfg_path: str, *args, **kwargs):
     cfg = OmegaConf.load(cfg_path)
     env = MettaGridEnv(cfg, *args, **kwargs)
     return env
 
+
```
sasmith: It's worth a comment to explain why this function needs to exist. E.g., why it needs to strip the paths.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +171,49 @@ def object_type_names(self):
     def close(self):
         pass
 
+class MettaGridEnvSet(MettaGridEnv):
```
sasmith, line 173: It seems like there's a design choice here of whether this should subclass `MettaGridEnv` or just be its own thing, and I think the reason for it subclassing is so you can pass a `MettaGridEnvSet` where you would otherwise pass a `MettaGridEnv` -- is that right?

But it seems definitely tricky to get this right, since it's now going to have properties like `map_height` that can be different from what we get if we call `make_env`.

I think it might be easier to reason about if this didn't subclass `MettaGridEnv`, and instead had something like `select_env` that would return a random `MettaGridEnv`.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +171,49 @@ def object_type_names(self):
...
+
+    def __init__(self, env_cfg: DictConfig, render_mode: str, buf=None, **kwargs):
+        self._env_cfgs = env_cfg.envs
+        self._num_agents_global = env_cfg.num_agents
+
+        env_cfg = self.select_env()
+
+        super().__init__(env_cfg, render_mode, buf, **kwargs)
```
sasmith, line 184: Calling `super` after you've set some instance variables has some code smell to it. The setting of `self._env_cfgs` and `self._num_agents_global` should ideally come after this.

sasmith (changes requested):
I think to review this better, I'd love to see how it's going to be used. Can we pair-or-similar?

mettagrid/mettagrid_env.py
```diff
@@ -1,27 +1,29 @@
 import copy
 import random
 from typing import Any, Dict
-
+import logging
```
sasmith: I think we don't need this now.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +170,48 @@ def object_type_names(self):
...
+    """
+
+    def __init__(self, env_cfg: DictConfig, render_mode: str, buf=None, **kwargs):
+        self._env_cfgs = env_cfg.envs
+        self._num_agents_global = env_cfg.num_agents
+
+        env_cfg = self.select_env()
+
+        super().__init__(env_cfg, render_mode, buf, **kwargs)
+        self._cfg_template = None
```
sasmith: This is interesting. Probably worth a comment like `we don't use this, and clear it to emphasize that fact`.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +170,48 @@ def object_type_names(self):
...
+
 def make_env_from_cfg(cfg_path: str, *args, **kwargs):
     cfg = OmegaConf.load(cfg_path)
     env = MettaGridEnv(cfg, *args, **kwargs)
     return env
 
+def config_from_path(config_path: str) -> DictConfig:
+    env_cfg = hydra.compose(config_name=config_path)
+
+    # need to strip the path-keys to have a valid config
```
sasmith: Can you expand this to say why? I think this could be
`When hydra loads a config, it "prefixes" the keys with the path of the config file. We don't want that prefix, so we remove it.`

mettagrid/mettagrid_env.py
```diff
@@ -1,27 +1,29 @@
...
-        self.make_env()
+        self.reset_env()
         self.should_reset = False
         self._renderer = None
 
         super().__init__(buf)
 
-    def make_env(self):
-        self._env_cfg = OmegaConf.create(copy.deepcopy(self._cfg_template))
+    def reset_env(self, env_cfg: DictConfig | None = None):
```
sasmith: I think we could do this a little differently, especially because having `reset_env` take `env_cfg` means that `MettaGridEnv.reset_env` and `MettaGridEnvSet.reset_env` now have different signatures.

Maybe there should be a `set_env` function that both versions of `reset_env` call?

Also note that this is copying the config, which shouldn't be needed in the case of `MettaGridEnvSet`, since in that case we reading new copies of the config (so we're already working with copies).

sasmith (changes requested):
Requesting changes to remove `logging` (high certainty, medium-low importance) and probably update `reset_env` (medium certainty, medium importance).

mettagrid/mettagrid_env.py
daphnedemekas: Thank you, that's helpful. was struggling with the wording here

mettagrid/mettagrid_env.py
daphnedemekas: Ok, i am slightly unsure about this logic but let me know what you think:

I have a set_env() function both in MettaGridEnv and MettaGridEnvSet

MettaGridEnv.set_env() just sets like this:

```
    def set_env(self):
        self._env_cfg = OmegaConf.create(copy.deepcopy(self._cfg_template))
        OmegaConf.resolve(self._env_cfg)
```

And MettaGridEnvSet.set_env() sets like this, replacing the select_env() function and just setting the sampled env as class variable

```
    def set_env(self):
        selected_env = random.choice(self._env_cfgs)
        env_cfg = config_from_path(selected_env)
        if self._num_agents_global != env_cfg.game.num_agents:
            raise ValueError("For MettaGridEnvSet, the number of agents must be the same for all environments. Global: {}, Env: {}".format(self._num_agents_global, env_cfg.game.num_agents))
        self._env_cfg = OmegaConf.create(env_cfg)
        OmegaConf.resolve(self._env_cfg)
```

Then in MettaGridEnv.reset() I first call set_env, and then call reset_env. So I no longer need MettaGridEnvSet.reset_env() either. 

I feel like the set_env and reset_env is a little confusing one after the other though:

```
        self.set_env()
        self.reset_env()
```

But maybe its fine

daphnedemekas:
@sasmith 
let me know if you're happy with the changes since yesterday

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +170,46 @@ def object_type_names(self):
...
+class MettaGridEnvSet(MettaGridEnv):
+    """
+    This is a wrapper around MettaGridEnv that allows for multiple environments to be used for training.
+    """
+
+    def __init__(self, env_cfg: DictConfig, probabilities: List[float] | None, render_mode: str, buf=None, **kwargs):
+        self._env_cfgs = env_cfg.envs
+        self._num_agents_global = env_cfg.num_agents
+        self._probabilities = probabilities
+        self._env_cfg = self._get_new_env_cfg()
```
sasmith, line 182: I think this shouldn't be needed, since it's handled in the `super().__init__`? Not a blocker, but nice to clean up before merging.

mettagrid/mettagrid_env.py
```diff
@@ -168,11 +170,46 @@ def object_type_names(self):
     def close(self):
         pass
 
+class MettaGridEnvSet(MettaGridEnv):
+    """
+    This is a wrapper around MettaGridEnv that allows for multiple environments to be used for training.
+    """
+
+    def __init__(self, env_cfg: DictConfig, probabilities: List[float] | None, render_mode: str, buf=None, **kwargs):
```
sasmith, line 178: Do you have an example of how probabilities can be defined at this level? E.g., I would have more expected the probabilities to show up in `env_cfg.envs`, but I'm hazy on what the hydra config will look like wrt this. Not a blocker.

sasmith (approved):
This looks good and clean. I think we may want to come back at some point and take another swing at getting the naming of things to be just right (e.g., `reset` vs `reset_env`), but I explicitly think we _don't_ want to do that now.

mettagrid/mettagrid_env.py
daphnedemekas: They show up in the multienv_mettagrid.yaml:

```
_target_: mettagrid.mettagrid_env.MettaGridEnvSet

num_agents: 24

game:
  num_agents: ${..num_agents}

envs:
  - /env/mettagrid/simple
  - /env/mettagrid/teams
  - /env/mettagrid/bases

probabilities:
  - 0.5
  - 0.25
  - 0.25

```

PR #73, opened 2025-04-07
New map generator

berekuk:
# Summary

<s>This is a draft, but I'm at the stage where I want to get some feedback on whether the team is ok with this approach.</s>

`mettagrid.map.mapgen.MapGen` class provides a new way to construct maps, based on scenes that can include other scenes recursively, and you select children areas to do that in bulk, all through YAML configs.

Example: "render a grid of 2x3 rooms; then add two mazes in two random rooms; add 1 agent in each room; add 1 mine and 2 generators in each room; add this handcoded ASCII map in a room that's not a maze".

`mapgen_demo.yaml` config, that can be viewed with `python -m tests.show_map game/map_builder=mapgen_demo game.num_agents=6`, should serve as a detailed summary of current capabilities.

Screenshot 2025-04-07 at 20 23 29

This PR includes the commits from #68, because I needed the `tests.show_map` script for development.

# Notes

## `configs/game/map_builder/` refactoring

I wanted to make configs reusable, so I've moved parts of old `configs/*.yaml` files to separate files.

This caused two problems:
1) If this PR is merged, the main Metta repo will break - it symlinks to only some configs, which I didn't realize initially.
2) In the current design, `game.map_builder` is tightly coupled with `game.num_agents`.

I'd argue that maybe `game.num_agents` shouldn't exist at all — the env constructor could export `map_agents` by counting the agents that are present on the map. But this is not a blocker for this PR.

For symlinks, I can revert the refactorings for older configs. Another thing that could be done is to `ln -s deps/mettagrid/configs configs/env/mettagrid`, and move extra mettagrid configs in Metta repo under `configs/env/mettagrid-extra`. 

## Demos for old -> new system

`configs/game/map_builder/` includes separate `maze.yaml` and `mapgen_maze.yaml` and similar pairs, to check that the new system supports all previous maps.

(I haven't converted cognitive_evals maps yet.)

The maps are not 100% exactly identical, but they're very close.

## Future plans

On my current roadmap:
- BSP tree maze generation (rougelike), since it's the most straightforward way to demo the new capabilities in a less boring environment than separate isolated rooms
  - I'm reimplementing this from scratch, btw, because the full scene implementation needs to export *areas* (rooms and corridors), so I couldn't use [urizen](https://github.com/vurmux/urizen)
- WaveFunctionCollapse integration
- Non-rectangular areas (probably with something like... rectangular area + a bitmask?)
  - Rationale/example: imagine that BSP scene exports "room" areas and "corridor" areas; then we want to populate corridors with items. If the corridor has turns then the rectangular area that covers the corridor could cover a corner of some unrelated room.
- Other scenes, such as:
  - "fill this area with walls of this shape, but make checks that it's still fully connected"
  - Layout scenes that allow you to target sub-areas like "left side of the map", or "top-left corner with 20% width/height"
- Built-in config flag for rotating scenes


berekuk:
BSP (roguelike) map generator, first pass.

Corridors sometimes stick to rooms, needs some more tuning.

Screenshot 2025-04-07 at 22 07 44

Note that placing agents is not the part of the generator code. This is done by a separate scene in config (`Random`).

berekuk:
A bit more complex example: "dungeon with 20 rooms, one room is for treasure, 3 other rooms have 2 agents each, and 2 more rooms are mazes".

Screenshot 2025-04-07 at 22 32 54

Full config:

```
_target_: mettagrid.map.mapgen.MapGen

width: 100
height: 120
border_width: 1

root:
  _target_: mettagrid.map.scenes.bsp.BSP
  rooms: 20
  min_room_size: 3
  min_room_size_ratio: 0.5
  max_room_size_ratio: 0.9

  children:
    - limit: 1
      lock: lock1
      scene:
        _target_: mettagrid.map.scenes.random.Random
        objects:
          generator: 3
          altar: 3

    - limit: 3
      lock: lock1
      scene:
        _target_: mettagrid.map.scenes.random.Random
        agents: 2

    - limit: 2
      lock: lock1
      scene:
        _target_: mettagrid.map.scenes.maze.MazeKruskal
        extra_space_empty: true
```

PS: this example also shows the limits of composability - in this case I got lucky, but in other generations the maze entrances were often blocked. This could be solved by registering the corridor entrance points, and then passing them to maze generators as parameters somehow, but it's not trivial.

berekuk:
First take at porting https://github.com/mxgmn/ConvChain/ (same author as WFC, but slightly simpler - there's less code to port).

Four different configs here, see https://github.com/Metta-AI/mettagrid/pull/73/commits/239fd68c3976357e165081d4f4b959ed2d181562#diff-66496193cb00bdb19f3b134184caf19df60b85eff1386d9b4785d818a2c052c4 for details.

Claude 3.7 one-shotted most of the algorithm conversion from C#, tbh I haven't even read it carefully yet.

- no guarantees of connectivity yet (WFC would be more reliable here, ConvChain is more random)
- it's kind of slow, heavily depends on N in NxN pattern; O(2^(N*N)), so N=4 is not viable
- `iterations` and `temperature` heavily affect the output, when I set `iterations` to be too large it deteriorates into an empty field (not sure why, I should look more into this). So I'm not sure it's possible to set these randomly and get a good result every time. It's possible to use random seeds and still get many maps when these parameters are tuned manually.

Screenshot 2025-04-08 at 20 11 32

sasmith:
I'm currently reviewing just at a looking-at-the-conversation level, since I think you're looking for high-level directional feedback, not a code review at the moment. Lmk if I'm wrong about that.
* This looks good directionally.
* I think we want non-rectangular rooms, beyond just corridors (although they're a good example. As concrete examples, we should be able to tile with triangles or hexagons. That said, I expect rectangles to be the dominant room type, so having rectangles that can fill the space without needing to be cleanly aligned with each other (which I think some of your examples are?) seems more important than different shapes.
  * I think your "non rectangular areas" future focus is good here. Maybe generic convex polygons? Concave would be nice as well, but I can see a lot of reasons why concave could be easier.
* Symlinks seem like something we can figure out.
* I agree that we shouldn't be declaring num_agents twice. It's implicitly defined in the map, and it's probably valuable to have this denormalized (so you don't need to read the whole map to see how many agents there are), but it doesn't need to exist separately from the map.
  * It could also be helpful to think of this mapgen as something that should work standalone from mettagrid. I think this should impact how we think about the mapgen's config.
  * For most games (e.g., starcraft), you can play with fewer players than the maximum the map allows. I don't think we need to allow that here -- we can just generate a different map. (I don't think this matters very much for what you're doing, but wanted to share my mental model.)
  * Generally, I'd nudge you to focus primarily on map shape, and only secondarily think about content. I.e., focus on the layout of walls, versus agent / resource placement. Like, I think we'll want to be able to say "there should be an agent in one of these three rooms", but I think this sort of thing it something we can solve on top of map generation.
* Nitpick regarding the corridors: I don't think we want them to be long, and I think we want them to (usually) be wider. In particular, agents spending a lot of time walking through corridors is going to be boring, and we also don't want agents to block each other in corridors of they're going opposite directions.
* The ConvChain maps look great and very organic. I hear you saying that they're expensive. As a rule of thumb, I think we should aim for map generation to cost less than 2% percent of whatever we're spending on training. So map generation taking "a few seconds" seems good, and taking "a few minutes" seems bad. If you're getting great results and it's too slow, that still seems worth checking in, and just highlighting that it's something we'll need to invest in making faster.

berekuk:
> I think you're looking for high-level directional feedback, not a code review at the moment

High-level is good. I'd like to get a more specific feedback on one particular question: does the config approach in [mapgen_demo.yaml](https://github.com/Metta-AI/mettagrid/pull/73/files#diff-da942bba339bb0f26641d4e1965f86ed0442016573fab1e1922306680bf783ff) look good to you? If yes, and I hope so, then it might be worth it to convert the remaining maps to use it quickly and retire the old `mettagrid.config.room.*` system (this shouldn't take me too long, I've converted most of things).

That would be a clear checkpoint for this PR. Or we could merge this PR first and then I'd do the remaining conversion later.

> I think we want non-rectangular rooms, beyond just corridors
> Nitpick regarding the corridors: I don't think we want them to be long, and I think we want them to (usually) be wider

Yes, agreed. In BSP maps, room sizes can be increased (`min_room_size_ratio`, `max_room_size_ratio`), and I could tune the corridor widths. With corridors in BSP, the checks that corridors don't merge with rooms in ugly way are a bit tricky, but that's solvable, and might not even be important.

> I think we'll want to be able to say "there should be an agent in one of these three rooms", but I think this sort of thing it something we can solve on top of map generation.

There's a dimension to this project that we've discussed on Discord, about hands-off approach (where map is generated based only on `size` and `complexity` parameters) vs finer control (where the semantics are described in configs).

The initial version of this PR was more focused on finer control, and one of the key ideas here is that the output of a scene is not just a grid, but a set of "areas". After I've switched more to hands-off, and started looking into things like ConvChain and WaveFunctionCollapse, I became less sure about this — when the map is generated with ConvChain, the generator doesn't know anything about the map's semantics.

Still, I think it could be neat to have the support for both approaches in the same system, so that specific eval maps could be built with the same system. Also, in some cases we could extract features from the generated map after the fact.

I have two big directions I want to try next:
1. WaveFunctionCollapse with many patterns - maybe it'll generate enough complexity that a single WFC scene is enough for all interesting maps.
2. Fractal scenes that are composed of modules, but the hierarchy and children scenes is determined randomly instead of written out explicitly in the config. This will require a few more types of scenes, some of which are "maze creators", other are "object creators", and others are "embellishers" (transform the maze, drop obstacles in free spaces, randomize walls), etc.

How (2) might work: the top-level MapGen module is parameterized with width/height and some complexity budget.
It randomly picks the children structure, then a random scene for each child, and each random scene does the same. The complexity budget is divided by children and passed down until it's exhausted.

(2) is clearly more complicated than (1) and hand-crafting any single map generator. At the same time, it could be very powerful and flexible. Right now I'm focusing more on WFC, and will decide if we need something more after we see the diversity we can get from WFC and ConvChain approaches.

> So map generation taking "a few seconds" seems good, and taking "a few minutes" seems bad.

Got it. We can also make the generation phase offline, and pick the maps from the pre-generated database for training, I think.

berekuk:
> It's implicitly defined in the map, and it's probably valuable to have this denormalized (so you don't need to read the whole map to see how many agents there are), but it doesn't need to exist separately from the map. 

Counting agents in the map sound like something that can be done very easily though?

We might want to know agent count in advance if map loading is slow, if we want to load policies and generate maps in parallel. Not sure how useful this is.

There are two levels to this question btw:
1. map _config_ drives the agent count
2. generated map drives the agent count

I'm mostly interested in (1) right now; this is because I constantly have to do `python -m tests.show_map game/map_builder=mapgen_convchain  game.num_agents=4`.

I can't put the num_agents in `game/map_builder/mapgen_convchain.yaml`, because it's higher in the hierarchy. And I can't make use of `num_agents` in the map generator in the general case, the `num_rooms: ${div:${..num_agents},6}` trick from the older `simple.yaml` config doesn't generalize.

Of course, maybe I should just move configs to `game/`, and put this there...
```
game:
  num_agents: ...
  map_builder: ...
```

At the same time, I think doing (2), i.e. counting agents generated by the builder after the fact, would be fine, and cleaner than the current approach. If you agree, I can send this as a separate smaller PR.

sasmith:
I don't completely follow what you're proposing, so I'll say how I'm thinking about things. I also don't think I have a model of "the one right way to do things", so I'm happy for your input on what you think would be more successful.

One-step process
There's a `map generator config` that gives input the the map generator. Amongst other things, this config says how many agents should be on the generated map. mettagrid never needs to look at the `map generator config`, it only looks at the map; so from that perspective, the number of agents is "set by the map".

Two-step process
There's an initial map generation stage that generates a map with no information about agents (or other artifacts on the map) -- it just generates terrain. There's a second stage that places artifacts on the map, and the configuration for this second stage includes a number of agents to place. Once again, mettagrid doesn't need to read the config -- it just reads the number of agents from the config.


However, in both cases, I would expect there's some higher-level process that's determining how many agents there should be, and what models should be used, and things like this. And I'd expect this higher level process to be controlling which maps are used in a way that controls how many agents should be placed on the map. E.g., it could be writing the second-stage map generation configs. I don't _think_ this matters from the point of view of map generation; but it could lead to some duplication of how we specify agents. E.g., there could reasonably be a part of the config that say "we want to train agents in a 1v1 setting, so we should generate maps with 2 agents", and then if the map generation generates maps with 3 agents, we'll hit some sort of exception, since we won't be able to correctly place the two agents as planned.

> At the same time, I think doing (2), i.e. counting agents generated by the builder after the fact, would be fine, and cleaner than the current approach. If you agree, I can send this as a separate smaller PR.

If this is straightforward, then yes, send me the PR. I think that'll clarify some confusion I have.

====

Overall, though, if we can think of this as a two step process, where we're first generating terrain and secondarily adding agents and other artifacts, then we can delay having to think about num agents. Would that help?

berekuk:
I think one-step or two-steps (or more! the new architecture in this PR is largely about an arbitrary number of steps) as mostly independent from the num_agents change.

Might be easier to give an example. This config produces 8 agents:

```
_target_: mettagrid.map.mapgen.MapGen

width: 50
height: 40

root:
  _target_: mettagrid.map.scenes.room_grid.RoomGrid
  rows: 2
  columns: 3
  children:
  # Add cylinders in two random rooms
  - limit: 2
    scene:
      _target_: mettagrid.map.scenes.inline_ascii.InlineAscii
      data: |
        WWWWWWWWWW
             A    
        WWWWWWWWWW
  # Add one agent in each room
  - scene:
      _target_: mettagrid.map.scenes.random.Random
      agents: 1
```

The number of agents is static and driven by the config. The config can only be interpreted in two ways:
1. By a user who reads the config and knows now scenes work
2. By producing the map and then counting the agents

It's possible to make scenes that would add a random number of agents, but that depends on the specific configuration. It's possible to write a configuration that makes the number of agents completely predictable (and still driven by MapGen).

Passing the agents in a top-down fashion in this approach won't work well in general, though: I'm aiming to make MapGen flexible to make the decisions on what to put in the map on each level of scene tree. This makes the system expressive while keeping the control over the process.

The question about num_agents for me is about whether we can always grab their number from the environment. My impression is that we can, and `num_agents` mostly serves as an assertion. (We can keep it as an assertion, even, but make it optional.)

I also want to highlight that this is **not a blocker**. This PR works with explicit `num_agents`, there's just a minor usability issue because of duplication. After looking at all places in the codebase that refer to `num_agents`, I think it's better if I do this after this PR is merged (some parts are tricky).

berekuk:
I've marked this PR as ready for review.

I could continue to build upon it, but it seems useful to get this in the main so that others could run trains with new maps.

There's one blocker here: [Mettagrid symlinks refactor](https://github.com/Metta-AI/metta/pull/124) PR **must be merged before this**. After that, there are no breaking changes here.

configs/game/map_builder/mapgen_demo.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+    # Scene will be instantiated for each matched area.
+    scene:
+      _target_: mettagrid.map.scenes.maze.MazeKruskal
+
+      # Scenes can be nested arbitrarily.
+      # MazeKruskal defines a 1x1 area at each of the four corners, so we can use these to place an agent and an altar.
+      children:
+      - where: # similar to SQL "WHERE" clause; selects areas based on their tags.
+          # Each corner is tagged, so we can select them individually.
```
sasmith, line 59: Maybe call out that the tagging is also specific to MazeKruskal?

configs/game/map_builder/mapgen_demo.yaml
```diff
@@ -0,0 +1,103 @@
...
+        scene:
+          _target_: mettagrid.map.scenes.inline_ascii.InlineAscii
+          data: a
+
+  - # No conditions, so all areas are selected.
+    scene:
+      _target_: mettagrid.map.scenes.random.Random
+
+      # Let's place some objects in each room.
+      # Note that for the maze rooms, Random scene will take existing walls into account.
```
sasmith, line 78: Does "take existing walls into account" mean that we won't place on existing walls? For a maze, it seems like this might block passage through the mine? (At the moment, I'm mostly saying "the documentation is good, and here are points you could clarify to make it better")

configs/game/map_builder/mapgen_demo.yaml
```diff
@@ -0,0 +1,103 @@
...
+      objects:
+        mine: 1
+        generator: 2
+
+  # What if we want to draw a custom scene in one room, but that room shouldn't have a maze? Let's use a lock.
+  #
+  # Lock will exclude the areas that were already selected by previous children that use the same lock name.
+  # So this guarantees that the custom map will not overlap with the maze.
+  - lock: lockname
```
sasmith, line 87: It seems like the most "pure" version of this is to have a set of locknames, and only select rooms whose current content doesn't intersect with our locknames. E.g.,

  - locks:
    - walls
    (place some walls)
  - locks:
    - generators
    (place some generators)
  - locks:
    - walls
    - generators
    (have a different type of child that places both walls and generators)

I was also dissatisfied with the "lock" / "locks" term when I thought of it as a noun; but when I think about it as a verb, I think it's correct.

configs/game/map_builder/mapgen_maze.yaml
```diff
@@ -0,0 +1,32 @@
...
+  columns: 2
+
+  children:
+  - scene:
+      _target_: mettagrid.map.scenes.maze.MazeKruskal
+
+      extra_space_empty: true
+
+      children:
+      - where: # similar to SQL "WHERE" clause; selects areas based on their tags.
```
sasmith: I don't think we need the comment, given the existence of the demo.

berekuk:
More scenes and features:

WFC is now much faster. 400x400 map takes 32 seconds (on M1 Pro), performance scales linearly. This is ~16x slower than most other generators like maze. I don't think I can optimize it significantly more without rewriting in C++.

Maze scene allows arbitrary values for room_size and wall_size:
Screenshot 2025-04-13 at 22 02 26

New MakeConnected scene that can be applied on top of any output to make it connected; example for 4 rooms grid -> convchain -> make_connected:
Screenshot 2025-04-13 at 21 58 11

The MakeConnected algorithm sometimes drills too many tunnels because I create tunnels from the largest connected component to all others, and doesn't check for duplicate work; finding the minimal subset of walls to remove is not trivial. But it does always succeed and it's guaranteed that it won't explode in complexity. Not sure if spending more time on it is the high priority.

Mirror scene that creates a sub-scene and reflects it horizontally, vertically or both.
Here's an example with 4 rooms for the same WFC pattern. First one is unmirrored, others show various mirror modes:
Screenshot 2025-04-13 at 22 12 33

configs/game/map_builder/mapgen_demo.yaml
```diff
@@ -0,0 +1,103 @@
+# This config shows the features of the mapgen system.
+_target_: mettagrid.map.mapgen.MapGen
+
+# With mapgen, top-level width and height must be known in advance. We don't do layout based on children components.
+# So this is not exactly the same as the old simple.yaml config, but it's very close.
+width: 50
+height: 40
+
+# Outer border width. Defaults to 6.
```
sasmith: Worth calling out that right now (at least for the root?) the border is outside of the width / height. This seems more convenient for the root, but it seems problematic / confusing for internal rooms. Seems easily changeable in the future, so not something we need to bike-shed on now.

mettagrid/map/mapgen.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self._grid = np.full(
+            (height + 2 * border_width, width + 2 * border_width), "empty", dtype="<U50"
+        )
+        self._grid[:border_width, :] = "wall"
+        self._grid[-border_width:, :] = "wall"
+        self._grid[:, :border_width] = "wall"
+        self._grid[:, -border_width:] = "wall"
+
+        # TODO - add outer wall borders, always useful
```
sasmith: As a bonus, this would make it easier to make borders be internal to rooms (since this particular border could be special and outside of the dimensions). (Still a risk of bike-shedding.)

mettagrid/map/node.py
```diff
@@ -0,0 +1,98 @@
+import random
+from typing import TypedDict
+import numpy as np
+import numpy.typing as npt
+
+
+class Area(TypedDict):
```
sasmith: Why a TypedDict, vs just an object or a dataclass?

mettagrid/map/node.py
```diff
@@ -0,0 +1,98 @@
...
+                    if match:
+                        selected_areas.append(area)
+        else:
+            selected_areas = areas
+
+        # Filter out locked areas.
+        lock = query.get("lock")
+        if lock:
+            if lock not in self._locks:
+                self._locks[lock] = []
```
sasmith: nit: feels like the value should be a Set()

sasmith (commented):
Sorry it's taking me so long to review. As you can see by the comments so far, I'm somewhat less than half way done. I'll try to prioritize it tomorrow.

mettagrid/map/node.py
berekuk: Oh right, that's more idiomatic.
(It's been a few years since the last time last programmed in Python, this is accidental TypeScript-ism, an attempt to define a type only and keep using dicts because those feel more lightweight. But that's not really true in Python.)

configs/game/map_builder/mapgen_demo.yaml
berekuk: Oh. Maybe I should've looked into existing object behaviors more, I mostly ignored the specifics and assumed that they're all swappable.

I think the right thing to do here is to make `MakeConnected` scene configurable, pass in object types that should be treated like walls. It's somewhat tricky though, because `MakeConnected` creates tunnels by turning walls into empty objects. If we sprinkle on objects and agents first, then dig, then a random number of agents could be deleted.

So the correct pattern (for generating a random well-formed map) is something like:
1. ...make any walls-only map
2. Random (objects)
3. MakeConnected (dig through walls and mines)
4. Random (agents)

configs/game/map_builder/mapgen_demo.yaml
berekuk: Yeah, I've considered multiple locks but couldn't think of any real-world examples where they'd be worth it yet. Seems easy to add later.

Very open to alternatives to "lock", couldn't find any.

configs/game/map_builder/mapgen_demo.yaml
berekuk: Yep, this is only for the root.

Lack of layout capabilities (where child scenes affect the size of child scenes) is the important limitation of this approach, so scenes don't have "width" and "height" at all.

Btw, I've updated the default width to 1 in the latest commits.

mettagrid/map/mapgen.py
berekuk: I've removed the comment, it was stale.

I don't think it's useful to have borders for most scenes, since we want them to be connected (with the exception of RoomGrid).

mettagrid/map/node.py
berekuk: changed, thanks

configs/game/map_builder/mapgen_demo.yaml
sasmith: The objects I'm talking about are things like mines and altars, which are permanent fixture (maybe can be destroyed based on map configurations, but not moved). We have recently introduced "blocks" which are movable walls, so we have that complication too.

I think there are a few options, in addition to what you suggested:
  * Replace walls with the objects. For this, only select walls that are bordered by at least one empty space.
    * This might be fine. It also might and up with crowded objects.
  * Add objects in spaces that are already "open". E.g., if the space is surrounded on all eight sides by empty squares, then the object won't be blocking any paths.
    * I somewhat prefer this. I think it'll make the most "natural" maps.
    * It requires that rooms be sufficiently large to hold the needed objects, but this seems more natural than needing to carve out walls in order to fit things.

berekuk: > Add objects in spaces that are already "open". E.g., if the space is surrounded on all eight sides by empty squares, then the object won't be blocking any paths. 

Don't we want agents to learn to interact with objects in corners though, and all other possible situations? (Example: cylinder cognitive_eval which intentionally places the altar in a narrow corridor)

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+from mettagrid.map.scene import Scene
+
+Direction = Literal["horizontal", "vertical"]
+
+
+class BSP(Scene):
+    def __init__(
+        self,
+        rooms: int,
+        min_room_size: int = 3,
```
sasmith: I'm generally apprehensive about scalar defaults in code. I.e., "False" and "None" are reasonable, but 3, 0.4, and 0.8 feel like they'd fit more naturally in a default config file. My real justification is "smell", but I think the reasoning behind that is that things like False and None are more unarguable defaults, whereas I could see people wanting to tweak the more continuous values, and this is the wrong place for it.

That said, this is also a nitpick, and shouldn't block the code being checked in.

(Also, maybe 3 is justifiable as we want to be able to navigate the room?)

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+        # [
+        #   layer1,
+        #   layer2, layer2,
+        #   layer3, layer3, layer3, layer3,
+        #   ...
+        # ]
+        zones = [Zone(0, 0, grid.shape[1], grid.shape[0])]
+
+        for _ in range(self._rooms - 1):  # split rooms-1 times
+            zone = zones[next_split_id]
```
sasmith: I think this could be simplified if we did `zone = zones.pop(0)`, so the list only contains zones we're using. This would have bad performance, which might not matter; but if you use a `deque` then it definitely won't matter.

mettagrid/map/scenes/bsp.py
sasmith: Can you add a comment about the expected behavior if we split to the point where rooms have less than `min_room_size` available to them? I would have expected an assertion, or for us to stop splitting. I think in practice, we keep splitting, and I think risk having rooms overwrite each other, and maybe have us try to write outside of the grid?

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+    └────────────┘
+
+    In this example, the surface is the set of . characters that can be approached from below (side="down").
+
+    The empty areas on the left and right are not part of the surface.
+
+    The code that uses the surface doesn't care what the surface is made of, it just needs to know which that can be approached.
+    """
+
+    def __init__(self, min_x: int, ys: list[int], side: Literal["up", "down"]):
```
sasmith: The meaning of `min_x` and `ys` wasn't clear to me. Probably worth some explanation, like `The surface here is made up of points [(min_x, ys[0]), (min_x + 1, ys[1]), ...]`.

Alternatively, could this just be a list of tuples?

Also, how hard would it be to make this omnidirectional, instead of just "up" and "down"? This could even be the whole perimeter of a (convex) shape, and which directions can be approached from each point on the perimeter.

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+            valid_xs.append(i)
+
+        x = random.choice(valid_xs)
+        return (x + self.min_x, self.ys[x])
+
+    @staticmethod
+    def from_zone(
+        grid: np.ndarray, zone: Zone, side: Literal["up", "down"]
+    ) -> "Surface":
+        # Scan the entire zone, starting from the top or bottom, and collect all the y values that are part of the surface.
```
sasmith, line 273: It's not clear to me why this is needed -- aren't Zones rectangular?

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+            length = -length
+            if direction == "horizontal":
+                start = (start[0] - length + 1, start[1])
+            else:
+                start = (start[0], start[1] - length + 1)
+
+        self.start = start
+        self.length = length
+
+    def transpose(self) -> "Line":
```
sasmith, line 334: Ah, it seems like this is needed because we only do vertical connections / surfaces. Or, rather, that we do horizontal by first transposing and then doing vertical, and then transposing back. This definitely seems worth a comment.

Mostly, my instinct when thinking about lines is to expect this function to rotate the line, and I find myself surprised to see the x and y for the line's start being switched and realize that it really is transposing the line.

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+
+    def __repr__(self):
+        return f"Line({self.direction}, {self.start}, {self.length})"
+
+
+def connect_surfaces(surface1: Surface, surface2: Surface):
+    """
+    Connect two surfaces with a corridor.
+
+    Assumes that the surfaces are adjacent and the surface1 is strictly above of surface2, i.e. all its positions are strictly smaller than all of surface2's positions.
```
sasmith, line 348: ... surfaces should have been transposed as needed to make this true.

mettagrid/map/scenes/convchain.py
```diff
@@ -0,0 +1,78 @@
...
+        children: list[Any] = [],
+    ):
+        super().__init__(children=children)
+        self._pattern = pattern
+        self._receptor_size = receptor_size
+
+        self._weights = ascii_to_weights_of_all_patterns(
+            self._pattern, self._receptor_size
+        )
+        # Ensure all weights are positive
```
sasmith, line 40: Can you justify this? I don't know what to expect from weights hardly at all, but I don't know why they can't be negative, or zero, or less that 0.1. Should that be done in `ascii_to_weights_of_all_patterns`?

Does it just come from the C# code and we're not sure?

mettagrid/map/scenes/bsp.py
```diff
@@ -0,0 +1,352 @@
...
+    lines = [
+        # Note: off-by-one errors here were quite annoying, be careful.
+        Line("vertical", start, turn_y - start[1] + 1),
+        Line("horizontal", (start[0], turn_y), end[0] - start[0]),
+        Line("vertical", end, turn_y - end[1] - 1),
+    ]
+    return lines
+
+
+def connect_rooms(room1: Zone, room2: Zone):
```
sasmith: Why not use `MakeConnected`?

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+
+        logger.debug("Populating distance to largest component")
+        distances_to_largest_component = self._distance_to_component(
+            node,
+            component_cells[largest_component_id],
+        )
+
+        # connect the largest component all other components
+        logger.info(f"Connecting {len(component_sizes)} components")
+        for component_id in range(len(component_sizes)):
```
sasmith: Seems like you want this to be `for component_id, cells in enumerate(component_cells):`

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+logger = logging.getLogger(__name__)
+
+
+class MakeConnected(Scene):
+    """
+    This scene makes the map connected by digging tunnels.
+
+    It does this by:
+    - Finding all the connected components
+    - Digging shortest tunnels from the largest component to all other components
```
sasmith, line 22: Consider if we want to join the two closest components instead? I guess that would make it n**2, but it seems like another natural way to join things.

Edit: I understand this better now, and think this might be something to go into the TODO that I'm proposing.

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+
+    TODO: This can result in some extra tunnels being dug.
```
sasmith, line 24: Surprising!

But also, I think for most maps I'd want some extra tunnels dug. It seems like the intended algorithm would result in a minimally connected graphs (with regards to inter-node connections). I think sometimes we'd want that, and sometimes we'd want more than minimally connected. Consider adding some option for extra edges, which we'd add by picking two random nodes and connecting them, or something.

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
+import logging
+import random
+from typing import Any, Tuple, Union
+
+import numpy as np
+from mettagrid.map.node import Node
+from mettagrid.map.scene import Scene
+
+directions = [(-1, 0), (0, 1), (1, 0), (0, -1)]
```
sasmith: prefer `DIRECTIONS`, to make it easier to identify as some sort of constant

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+        self,
+        node: Node,
+        component_cells: list[Tuple[int, int]],
+    ):
+        height, width = node.grid.shape
+        # find the distance from the component to all other cells (ignoring the walls - used for finding the optimal tunnels)
+        distances = np.full((height, width), np.inf)
+        queue = []
+        for cell in component_cells:
+            distances[cell[0], cell[1]] = 0
```
sasmith: If you want to be cute, consider `distances[*cell] = 0`

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+            return
+
+        # find the largest component
+        largest_component_id = max(
+            range(len(component_sizes)), key=component_sizes.__getitem__
+        )
+        logger.debug(f"Largest component: {largest_component_id}")
+
+        logger.debug("Populating distance to largest component")
+        distances_to_largest_component = self._distance_to_component(
```
sasmith, line 51: Ah! I expected this to work recursively, where once we connected something, we would then consider the newly connected component to be part of the largest component, and figure out the next connection based on this (so, e.g., the potential set of connection points would be increased). But we don't. Instead we're going to keep connecting things to the original largest component.

This seems fine, but I think I'd want to TODO about iterating on this. I see that performance could get really bad here if we need to repeatedly crawl the whole grid, but I expect we could make this more efficient by doing things like only considering exterior points and such.

Also, I think we should be more explicit about this being the logic (that everything connects to the originally largest component, vs it being iterative).

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+        # connect the largest component all other components
+        logger.info(f"Connecting {len(component_sizes)} components")
+        for component_id in range(len(component_sizes)):
+            if component_id == largest_component_id:
+                continue
+
+            # find the cell that's closest to the largest component
+            min_distance = np.inf
+            min_distance_cell: Union[Tuple[int, int], None] = None
+            for cell in component_cells[component_id]:
```
sasmith: min_distance_cell = min(cells, key: lambda c: distances_to_largest_component[*c])

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+            if min_distance_cell is None:
+                raise ValueError("No cell found for component")
+
+            # connect the cell to the largest component by digging a tunnel based on the shortest path
+            current_cell = min_distance_cell
+            while distances_to_largest_component[current_cell[0], current_cell[1]] > 0:
+                # Find the neighbor with the smallest distance to the largest component
+                y, x = current_cell
+                # Find all neighbors with the smallest distance to the largest component
+                candidates = []
```
sasmith: With every step, the distance should go down by 1. So we can probably clean this up a bit by having the candidates be all cells that are 1 closer than the current distance.

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+                    continue
+
+                components_cells.append([])
+                queue = [(y, x)]
+                i = 0
+                while i < len(queue):
+                    y, x = queue[i]
+                    i += 1
+                    if components[y, x] != -1:
+                        continue
```
sasmith, line 124: Could have an assertion that `components[y, x] == component_id`.

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+                    continue
+
+                # already visited
+                if components[y, x] != -1:
+                    continue
+
+                components_cells.append([])
+                queue = [(y, x)]
+                i = 0
+                while i < len(queue):
```
sasmith, line 120: This is fine as is, but you could consider using a stack here, and not needing `i`.

mettagrid/map/scenes/make_connected.py
```diff
@@ -0,0 +1,167 @@
...
+        height, width = node.grid.shape
+        # find the distance from the component to all other cells (ignoring the walls - used for finding the optimal tunnels)
+        distances = np.full((height, width), np.inf)
+        queue = []
+        for cell in component_cells:
+            distances[cell[0], cell[1]] = 0
+            queue.append(cell)
+
+        i = 0
+        while i < len(queue):
```
sasmith, line 158: Same as above, I think this could be a stack and then not need `i`. Or if you really wanted to make the point that order doesn't matter, it could be a set. So

```
to_visit = set(...)
while to_visit:
  y, x = to_visit.pop()
  ...
```

sasmith (commented):
Here's another chunk. I'm at 36/46 files.

mettagrid/map/scenes/convchain.py
berekuk: This is justified in (3) here: https://github.com/mxgmn/ConvChain/

> For the speed of convergence, it's convenient for weights of all patterns to be nonzero. So let's redefine the weight(P) of a pattern P to be the number of patterns P in the input if that number is more than zero and some small number eps otherwise, 0 < eps < 1.

But also tbh I don't really understand how ConvChain works yet... I've converted it from C# with Claude and verified that it matches the algorithm, but didn't have time to figure out what's going on, because I tried to cover as much as possible in a few days, so I've switched to trying WFC instead (which I do understand).

I'll get back to it soon.

mettagrid/map/scenes/bsp.py
berekuk: `MakeConnected` is more sloppy with its tunnels currently.

(Also, I wrote BSP first; I tried MakeConnected + BSP without corridors after I realized that this is a possibility, and it was not good enough, i.e. didn't look like a typical rougelike map)

mettagrid/map/scenes/random.py
```diff
@@ -0,0 +1,69 @@
...
+        assert len(symbols) <= empty_count, (
+            f"Too many objects for available empty cells: {len(symbols)} > {empty_count}"
+        )
+
+        # Shuffle the symbols
+        symbols = np.array(symbols).astype(str)
+        self._rng.shuffle(symbols)
+
+        # Place objects only in empty cells
+        if len(symbols) > 0:
```
sasmith: Feels like this check should be higher. But mostly, the code structure makes it look like this branch is the uncommon case; whereas the branch is the strongly-expected default case.

I.e., maybe add `if not symbols: return` up at line 48?

mettagrid/map/scenes/random.py
```diff
@@ -0,0 +1,69 @@
...
+        objects: DictConfig | dict = {},
+        agents: int | DictConfig = 0,
+        seed: Optional[int] = None,
+    ):
+        super().__init__()
+        self._rng = np.random.default_rng(seed)
+        self._objects = objects
+        self._agents = agents
+
+    def _render(self, node: Node):
```
sasmith, line 29: This makes impressive use of numpy functionality to make the random placement clean. I assume it's performant despite the re-formatting, since numpy should be fast.

mettagrid/map/scenes/room_grid.py
```diff
@@ -0,0 +1,62 @@
...
+            node.width - self._border_width * (self._columns - 1)
+        ) // self._columns
+        room_height = (
+            node.height - self._border_width * (self._rows - 1)
+        ) // self._rows
+
+        # fill entire node.grid with walls
+        node.grid[:] = self._border_object
+
+        for x in range(node.width):
```
sasmith: This looks like cruft from a previous version?

mettagrid/map/scenes/wfc.py
sasmith, line 1: Reviewed. No ability to leave helpful comments, except to note that this is pretty indecipherable without going to read documentation on the original. I'm torn about whether there should be more comments here, or just an expectation that people will follow the link if needed. Definitely feel free to add any comments that you think would be leveraged -- e.g., about usage.

mettagrid/map/utils/pattern.py
```diff
@@ -0,0 +1,184 @@
...
+import numpy.typing as npt
+
+Symmetry = Literal["all", "horizontal", "none"]
+
+
+def parse_ascii_into_grid(source: str) -> npt.NDArray[np.bool_]:
+    """
+    Parse an ascii string into a numpy ndarray of booleans.
+
+    The string must be enclosed in | characters, and must be composed of # and space characters.
```
sasmith, line 12: ... \# will be treated as walls, and spaces as empty.

mettagrid/map/utils/pattern.py
```diff
@@ -0,0 +1,184 @@
...
+    grid: npt.NDArray[np.bool_] = np.zeros((height, width), dtype=bool)
+    for y, line in enumerate(lines):
+        for x, char in enumerate(line):
+            grid[y, x] = char == "#"
+
+    return grid
+
+
+def ascii_to_patterns_with_counts(
+    source: str, n: int, periodic: bool = False, symmetry: Symmetry = "none"
```
sasmith: I think this is another place where we don't need defaults -- I think the defaults will all come in at a higher level.

mettagrid/map/utils/pattern.py
```diff
@@ -0,0 +1,184 @@
...
+            p1 = p0.reflected()
+            p2 = p0.rotated()
+            p3 = p2.reflected()
+            p4 = p2.rotated()
+            p5 = p4.reflected()
+            p6 = p4.rotated()
+            p7 = p6.reflected()
+            return [p0, p1, p2, p3, p4, p5, p6, p7]
+        elif symmetry == "horizontal":
+            return [self, self.reflected()]
```
sasmith, line 144: This seems fine, but I'm not sure of a case where we'd want only horizontal symmetry. Do you have an example? Mostly noting that you've chosen a handful of symmetries here, but not all (e.g., rotational), and I'm curious if you have a specific reasoning.

mettagrid/map/scenes/wfc.py
```diff
@@ -0,0 +1,293 @@
...
+            np.log(self._sum_of_weights)
+            - self._sum_of_weight_log_weights / self._sum_of_weights
+        )
+
+        self._next_node_heuristic = next_node_heuristic
+        self._attempts = attempts
+
+        self._fill_propagator()
+
+    def _fill_propagator(self):
```
sasmith, line 86: "Here we pre-calculate pattern compatibility, and we'll use this to check the compatibility of potential patterns going forward. An alternative could be to write (/ flatten) chosen patterns to the grid, and then check potential patterns against the grid; but this is easier."

sasmith (changes requested):
Some additional feedback. Nothing that should be blocking, but I'm marking as requesting changes due to the number of comments, and my inability to keep the whole diff in my head at once.

mettagrid/map/scenes/bsp.py
berekuk: I needed non-leaf zones too, for corridors, later in the code.

I've refactored this code  - there's a `BSPTree` class now that encapsulates most of the logic, and I'm reusing it in the new `BSPLayout` scene that's used in `Auto` scene that I got in after your review.

mettagrid/map/scenes/bsp.py
berekuk: > Alternatively, could this just be a list of tuples?

In current semantics, the surface is always a continuous chunk, so went with the minimal representation (not just for memory saving, but for the sake of "make impossible states unrepresentable").

I could add a function that returns tuples, but it seems more safe to store the source of truth in the minimal form.

> Also, how hard would it be to make this omnidirectional, instead of just "up" and "down"?

The direction is not just the Surface class limitation, it's part of its contract: the code that builds corridors relies on the direction.

Btw, this is all very specific to BSP, and it's not clear how much BSP is going to be used in practice (it seems like WFC and noise generators are more flexible). I'm making use of BSPLayout though and expect that one to stay.

mettagrid/map/scenes/bsp.py
berekuk: The zone is rectangular, but the shape inside the zone can be pretty complex.

The reason why I needed all this is pretty technical and I'm not sure it's worth improving it. There might be an easier way, or maybe it's extra complexity that could be ignored.

Consider this maze:
```
################
#             ##
#    #####    ##
#    #####    ##
################
################
################
####       #####
####       #####
################
```

There are three rooms here, two on the top that are already connected, and one on the bottom.

[One of the guides that I followed](https://www.roguebasin.com/index.php/Basic_BSP_Dungeon_generation) said "we can connect two sub-regions with a link either between two rooms, or a corridor and a room or two corridors."

So I needed to connect the bottom room with either the left room, right room or the corridor. And the top structure could be more complicated than this example (the corridor could have turns, there could be more rooms and more corridors, etc.)

The most annoying special case is that corridors can sometimes graze against the space of an adjacent room, like this:
```
################
#             ##
#    C####    ##
#    C####    ##
#####C##########
#####C##########
#####C##########
####       #####
####       #####
################
```

Ideally, we want to avoid that. The surface code is supposed to do this, but I'm not sure it handles all cases perfectly. I could do something more clever like pathfinding, not sure if it's worth extra effort though.

mettagrid/map/scenes/bsp.py
berekuk: I've added some comments.

mettagrid/map/scenes/make_connected.py
berekuk: Yes. LLMs suggested minimal spanning tree, but then naively I'd need to precalculate the minimal distance between each pair of components; I couldn't figure out how to make it definitely fast on the spot (component_count * map_size is still slow, what if we have a lot of components; I can't make any assumptions about the underlying structure, it could be lots of isolated 1x1 cells in the worst case, or weird concave shapes). Seemed solvable, but this was good enough for my goals, for now, didn't want to complicate things prematurely.

At the same time, it's annoyingly not good enough for BSP, or for mazes. (theoretically we could do RoomGrid of 1x1 rooms -> MakeConnected and get a maze... but not with this approach; this approach destroys most of the map).

I'll add a todo note, and spend some more time thinking if there's some easy way to do this that I've missed.

mettagrid/map/scenes/make_connected.py
berekuk: The order does matter! We want BFS in layers (first process everything at distance 1, then distance 2, etc.), otherwise the distances won't be computed correctly.

I think it doesn't matter in `_make_components`, but probably easier to use the same approach in both cases.

(For a bit of context, I did a lot of Codeforces contests in the past, and found this pattern quite robust for BFS, compared to the stack.)

mettagrid/map/scenes/random.py
berekuk: IIRC the first half of it is old code and the second half was one-shotted by Claude, actually :)

But yeah, looks like it should be quite fast.

mettagrid/map/scenes/wfc.py
berekuk: Yeah... 4d arrays for propagator and compatible are particularly hard to explain even in comments, I don't see a great way for people to grok this without careful reading through both the documentation and the source (which is what I did with the original C#, with some help from o1-pro to check that I haven't messed up the indices somewhere).

If I get back to this code later, e.g. to make it more flexible for some special case that we need, I'll try to add more comments and tests.

mettagrid/map/utils/pattern.py
berekuk: I ported these from WFC and ConvChain, mostly (original is [here](https://github.com/mxgmn/WaveFunctionCollapse/blob/a088882ea7261ca11f7e02b58b12a870d809e8a4/OverlappingModel.cs#L53) if you're curious).

The reasoning is:
- I wasn't sure what I would need
- I wanted to try the examples from WFC repo to see if my outputs are similar (still haven't done that in a systematic way; you can see their examples [here](https://github.com/mxgmn/WaveFunctionCollapse/blob/master/samples.xml); symmetry=2 is horizontal)
- symmetry="all" is definitely useful in some cases; and "horizontal" is a subset of that, so it was trivial to add both

mettagrid/map/scenes/wfc.py
berekuk: > An alternative could be to write (/ flatten) chosen patterns to the grid, and then check potential patterns against the grid; but this is easier.

I'm not sure what you mean by this:
- the main purpose of the propagator is performance optimization
- I don't see how we could write patterns to the grid without it; the wave function is expressed in pattern ids. these could be decoded to patterns and compared on each step, but there's no grid to write them on (since we maintain the list of possibilities that's narrowed incrementally)

Rephrased so:
> Here we pre-calculate pattern compatibility, and we'll use this to check the compatibility of potential patterns going forward, when we exclude some possibility from the wave function and want to propagate the change.

PS: General thought on comments, for this particular case: it's hard to guess the level of competence of people who would read this code later. The range of algorithmic competence among engineers is wide, from "this is obvious" to "ugh I beat my head against this code for a few days before I got it" (pre-LLMs and pre-Codeforces me would be in this latter category). LLMs might have an advantage here, because you can ask it specific questions.

mettagrid/map/scenes/make_connected.py
berekuk: I've rewritten this as `visited` (with bools), I never actually used indices.

Btw, I've just fixed a dumb bug here: `y` and `x` were overwriting outer `y` and `x`, so some rooms were not connected. (implicit variable declarations might be my least favorite Python feature...)

I've spotted this on large maps by accident; added an assertion at the end just to be safe: `assert len(self._make_components(node)) == 1`.

berekuk:
I've responded to everything in your comments; please say if you think I pushed back on something you cared about.
 
Added some more new code here, mostly related to Auto scenes.

Remerged with fresh main.

There's still more work to be done, but generally I think this is in good shape, and could be used as is for training with the latest Auto improvements.

(I still need to do WFC-patterns-stored-in-files and more WFC patterns in general, I remember; those don't seem like a blocker to me though)

configs/game/map_builder/mapgen_auto.yaml
```diff
@@ -0,0 +1,173 @@
...
+      # temple: ${uniform:1,200,40}
+      # TODO - blocks?
+
+    # The percentage of objects to generate in each area.
+    # The percentage will be sampled from the distribution for each room separately.
+    room_objects:
+      mine:
+        distribution_type: uniform
+        min: 0.001
+        max: 0.03
```
berekuk, line 35: Sorry for the inconsistency here vs `min_*` fields in other parts of this config! I started with naive min/max, and then implemented these more generic "distribution_types" for objects in rooms specifically, because I wanted to clearly see that distributions are clearly different, and uniform was not diverse enough.

I'd like to use these distribution type objects everywhere, eventually, but want to get the confirmation first that the pattern looks ok (semi-expect someone to push back in the direction of resolvers because of the need for sweeps or something).

sasmith: No strong opinion yet, and moreover, I'd like to be iterating from checked in code vs iterating a very large diff that's adding a bunch of other stuff we want. So I'd put this in the bucket of "fine to experiment, and expect we should follow up here sort of soon once we've played with it". That said, one of the very nice properties of having interpolations like `${lognormal(0.0001, 0.03, 0.05)}` is that we can use them anywhere in the config.

mettagrid/map/scenes/make_connected.py
sasmith: Ahh. You're right. If you don't do the right order, you could turn a corner and think that a close point is actually far. Got it.

mettagrid/map/utils/random.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if low >= high:
+        raise ValueError("Low value must be less than high value")
+    if low <= 0:
+        raise ValueError("Low value must be above 0")
+
+    # Default to 90% probability, for now
+    probability = 0.9
+
+    if probability <= 0 or probability >= 1:
```
sasmith, line 27: Perfectly fine, but Python also allows sugar like `if not 0 < probability < 1: ...`

Also a little funny to have the assertion following a constant, but I see that it's nice for future proofing.

mettagrid/map/utils/random.py
```diff
@@ -0,0 +1,52 @@
...
+        raise ValueError("Probability must be in (0, 1) interval")
+
+    log_low = np.log(low)
+    log_high = np.log(high)
+
+    # Calculate normalized sigmas using the inverse of the normal CDF
+    normalized_sigmas = stats.norm.ppf(1 - (1 - probability) / 2)
```
sasmith, line 34: I'm not doing a careful review of the math (partially because of the size of the diff). lmk if you think I should.

sasmith (approved):

configs/game/map_builder/mapgen_auto.yaml
berekuk: I wrote a longer argument why I prefer to avoid resolvers for this on Discord.

(short version: I'd have to call `OmegaConf.resolve` from specific places in map code to do this, and I might as well use arrays or dicts and call `my_resolve` instead)

> I'd like to be iterating from checked in code vs iterating a very large diff that's adding a bunch of other stuff we want

Same :)

mettagrid/map/utils/random.py
berekuk: For context: this was translated from https://github.com/quantified-uncertainty/squiggle/blob/87541969f4b359a9f3d84a8d48cdf9dd640dbfd5/packages/squiggle-lang/src/dists/SymbolicDist/Lognormal.ts#L78 (which I owned at my previous job) by Claude.

mettagrid/map/utils/random.py
berekuk: The original code I linked above was well-tested, so it should be fine.

PR #75, opened 2025-04-08
Remove Converter subclasses

sasmith:
The subclasses were mostly superfluous. We've now made them fully superfluous and removed them. We're still special-casing objects in mettagrid.pyx.

smoke-tested this by training locally.


daveey (approved):

sasmith:
There's something funny here where I had a stacked diff, and so I had PRs to merge B -> A -> main; and since you and @daphnedemekas merged the PRs yourself, we ended up with A merge to main, and B merge to A (versus getting A merge to main and then B merged to main). Since main isn't protected, it's easy for me to just do this merging myself now, but I think this is another thing that something like graphite could help with.

sasmith:
Also, you merged this instead of squash-merging. This is pretty confusing to me since the repo should be configured to only allow squash merging, and I write my commit messages with the assumption that we're just doing squash-merging. I.e., the commit messages for my intermediate commits are bad, whereas the message for my PRs are good. @daphnedemekas did the same on my other PR.

I'll make a task to follow up with y'all to see what the UI looks like for you to see how this is happening.

PR #76, opened 2025-04-10
George navigation training

georgejwdeane:
This PR includes changes from the training-suite PR, so waiting on merge for that


mettagrid/mettagrid_env.py
```diff
@@ -1,15 +1,17 @@
...
 from typing import Any, Dict
-
+import logging
 import gymnasium as gym
 import hydra
 import numpy as np
 import pufferlib
 from omegaconf import OmegaConf, DictConfig
-
+from util.config import config_from_path
```
berekuk: This won't work when mettagrid is used without metta.

daphnedemekas (approved):

PR #77, opened 2025-04-11
render_ascii for all objects; move ascii encoding to constants

berekuk:
Without this, `c_env.render` crashed on objects with id>5.

### DIFF
--- a/mettagrid/objects/constants.hpp
+++ b/mettagrid/objects/constants.hpp
@@ -53,6 +53,20 @@ std::vector<std::string> ObjectTypeNames = {
     "converter"
 };
 
+std::vector<std::string> ObjectTypeAscii = {
+    "A",
+    "#",
+    "g",
+    "c",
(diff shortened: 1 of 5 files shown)


mettagrid/objects/constants.hpp
```diff
@@ -53,6 +53,20 @@ std::vector<std::string> ObjectTypeNames = {
...
+    "#",
+    "g",
+    "c",
+    "a",
+    "r",
+    "l",
+    "b",
+    "f",
+    "t",
+    "v"
```
berekuk, line 67: Btw, it would be nice to get these in sync with the encoding used in ascii map files (those use `W` for walls, etc.).

sasmith (approved):
Looks good. We're directionally moving to have more object definition in configs and less in code. Ideally this will take ObjectType / ObjectTypeNames / ObjectTypeAscii out of code.

PR #79, opened 2025-04-14
Terrain generation

Dmitry273:
Terrain generation based on simplex noise and guided by arbitrary sampling functions + 2 scenarios for cognitive evals: maze and haul + small ASCII mazedrawing script


Dmitry273:
3r symmetry off-center rotation
5r symmetry rotation+
fn4 P from 0 01 to ~23
fn4_2 Pf0 01t23 exp + fn5_2 Mf0t1 lin
rotating parabolas f0t0 5

Examples of various sampling functions with time parameter for terrain generation

Dmitry273:
14 consecutive generations with fn7 calibrated layer integrated through yaml file

Dmitry273:
16 consecutive generations with new calibrated yaml file

mettagrid/config/room/cognitive_evals/mazedrawing.py
```diff
@@ -0,0 +1,63 @@
...
+    "converter": "\033[0;34mc\033[0m", # draws in blue
+    "generator": "\033[1;32mg\033[0m", # draws in light green
+    "mine": "\033[0;35mm\033[0m", # draws in purple
+    "wall": "W",
+    "empty": " ",
+    "block": "b\033[0m",
+    "lasery": "L\033[0m",
+}
+register_resolvers()
+config = OmegaConf.load('/home/catnee/mettagrid/configs/game/map_builder/mapgen_terrain_symmetry.yaml')
```
berekuk: Your personal home dir, won't work for anyone else.

I'd prefer that we fold this script into my `tests/mapgen.py`; it can render maps and show them in the standard mettagrid environment.

It can be argued that initializing the environment is overkill when we just want to generate the map. But even so, I think it's better to improve on the mapgen script (which has other features such as saving the maps as files for mettamap viewer, or different render modes) than to maintain two different scripts.

I'll also flag that loading configs from `./configs/game/map_builder` might fail if you use `OmegaConf` directly, instead of hydra. Those configs could use defaults lists which is the Hydra-only feature.

mettagrid/config/room/cognitive_evals/terraingen.py
berekuk, line 1: Is this file useful now that you've converted it to `mettagrid.map.scenes.terraingen`? I see that they're similar but not identical, I'm confused how they relate.

mettagrid/map/scenes/terraingen.py
```diff
@@ -0,0 +1,215 @@
...
+                                                                                                              # the faster noise is sampled: fn(x,y) ~ fn(x+1,y) in some region the less changes will be in this region per pixel, noise will look zoomed in
+                                                                                                              # the slower noise is sampled: fn(x,y) !=fn(x+1,y) in some region the more changes will be in this region per pixel, noise will look zoomed out
+        terrain = (terrain + 1)/2 # changes range from [-1,1] to [0,1]
+        terrain = terrain.reshape((self._height,self._width))
+        terrain = terrain.transpose()
+        terrain = terrain**layer.saturation # saturates pattern with walls. Helpful since base noise is balanced to be 50/50. Saturation 0 makes neutral terrain with 0 walls, saturation >100 fills everything with walls.
+
+        return terrain
+    
+    def fix_map(self, room: np.ndarray) -> None:
```
berekuk: Can you try adding `mettagrid.map.scenes.make_connected.MakeConnected` scene instead of this and see if it works better?

It has some limitations but it should be very fast. I'd prefer we have one canonical (and reusable) implementation of connectivity algorithm.

mettagrid/map/scenes/terraingen.py
```diff
@@ -0,0 +1,215 @@
...
+    xi, yi = xi * octave[0], yi * octave[1]
+ 
+    return (xi, yi)
+
+class Layer(NamedTuple):
+    fn: 'function'
+    saturation: float
+    params: dict[str,Any]
+
+class TerrainGen(Scene):
```
berekuk: Rename to `OpenSimplex` or something? `TerrainGen` is very generic.

mettagrid/map/scenes/terraingen.py
```diff
@@ -0,0 +1,215 @@
...
+        self.cutoff = cutoff
+        self._rng = np.random.default_rng(seed)
+        self.layers = [Layer(hydra.utils.get_method(x.fn), x.saturation, x.params) for x in layers if isinstance(x.fn, str)]
+        
+
+
+    def _render(self, node: Node):
+        grid = node.grid
+        self._width = node.width
+        self._height = node.height
```
berekuk: `Scene` class isn't supposed to have node-specific state.

I.e., "Scene" is a configuration that can render things into nodes; "Node" is a specific grid. The idea is that configuration can be reused, so Scenes are mostly dataclasses that can do some pre-initialization if necessary, cache things, etc. One scene can render onto multiple nodes from the same scene object.

So I recommend you pass the `node` or `node.grid` around explicitly.

(I hope to revisit Scene/Node APIs soon to avoid this issue but haven't completely figured out how exactly. Sorry for the confusion.)

mettagrid/map/scenes/terraingen.py
```diff
@@ -0,0 +1,215 @@
...
+        
+        terrain = np.ones(shape=(self._width,self._height)) # template neutral terrain of ones to prevent errors
```
berekuk: `shape=grid.shape`, and then you won't have to transpose the terrain in the end because it would have the `height, width` order.

.gitignore
```diff
@@ -17,3 +17,5 @@ player/dist
 player/node_modules
 /multirun
 .coverage
+poetry.lock
```
sasmith, line 20: Can you say more? By default, I don't think we want to ignore this -- I don't think we're using poetry.lock, but on principle, we should be.

.gitignore
```diff
@@ -17,3 +17,5 @@ player/dist
...
+/giffactory
```
sasmith: Also not sure on this one? Maybe this should go in `~/.gitignore`?

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+
+import math
+import time
+'''
+This file contains a set of functions that can be used to generate different types of noise patterns for terrain generation.
+The functions are designed to be used with the OpenSimplex noise generator, and they can be combined in various ways to create complex terrain features.
+The functions take in parameters such as x and y coordinates, width and height of the terrain, and various other parameters that control the noise generation process.
+'''
+
+def fn0(x,y, width, height,
```
sasmith: These functions should be less mysterious. This would include:
* Giving them more descriptive names
* Providing some sort of documentation as to what they add to the output, and how their parameters impact this. (e.g., "this  makes the noise more 'swirly'. The direction and tightness is controlled by foo. Bar controls how consistent the swirliness is each rotation.")

Generally I'd prefer for defaults to be in configuration rather than function definitions (defaults values like `None` are typically better in code than values like `0.1`, but it's fuzzy). It's hard for me to be sure what makes sense in this case, but if we're going to have constants in code, we should have some documentation around why these are appropriate defaults.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+    xc = 0.0
+    yc = 0.0
+    xi, yi = ((cs*(x-(0.5+xc)*width) + sn*(y-(0.5+yc)*height))+(0.5+xc)*height), ((-sn*(x-(0.5+xc)*width) + cs*(y-(0.5+yc)*height))+(0.5+yc)*height)
+
+    xi = (xi-(0.5+xc)*width)*zoom**2 /ell
+    yi = (yi-(0.5+yc)*height)*zoom**2 *ell
+
+    return (xi, yi)
+
+def fn1(x,y, width, height, **args) -> tuple[float, float]:
```
sasmith: Generally prefer not to check in unused code. If we're going to check in unused code, it should come with the same documentation as anything else, and also some explanation of why we're checking it in even though it's unused (e.g., "This is an experimental alternative to foo. Checking it in so people can compare to two. Eventually we should delete one of them.")

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+from mettagrid.map.node import Node
+from mettagrid.map.utils.random import MaybeSeed
+from collections import deque, defaultdict
+from queue import PriorityQueue
+
+import math
+import time
+'''
+This file contains a set of functions that can be used to generate different types of noise patterns for terrain generation.
+The functions are designed to be used with the OpenSimplex noise generator, and they can be combined in various ways to create complex terrain features.
```
sasmith: (My opinion here may change based on documentation changes elsewhere in the file)

This should come with examples of how the functions can be used / composed. Pointing to example yaml files may suffice.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+
+def fn2(x,y, width, height,
+        zoom:float = 0.1,           # global scaling
+        ell:float = 1.5,            # ellipticity
+        t:float = 0.25,             # angle of rotation in 2*pi radians
+        P:float = 2.0,              # thickness of the spiral
+        xc:float = 0.0,             # x off-center
+        yc:float = 0.0,             # y off-center
+        ) -> tuple[float, float]:
+    # function used in "spiral" generator
```
sasmith: This is a helpful breadcrumb. I think a lot of the documentation I'm looking for is "expand this".

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+
+    EMPTY, WALL = "empty", "wall"
+
+    def __init__(
+        self,
+        room_size: int = 1,
+        wall_size: int = 1,
+        seed: MaybeSeed = None,
+        children: list[Any] = [],
+        layers: list[Layer] = [], # layers dictate how generated noise is sampled and how the end result will look
```
sasmith: This is a helpful breadcrumb, but (as per all my other comments), I think this should be expanded. Like, _how_ do they dictate how the noise is generated? How does this interaction with `children`?

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+    saturation: float
+    params: dict[str,Any]
+
+class SimplexSampler(Scene):
+
+    EMPTY, WALL = "empty", "wall"
+
+    def __init__(
+        self,
+        room_size: int = 1,
```
sasmith: As per above, I'd like these parameters to be documented as well. Something like `seed` has a pretty obvious usage, and so probably need very little, assuming it's doing the obvious thing. `cutoff` is much less obvious. Others are in the middle.

sasmith (changes requested):
I started by reviewing mettagrid/map/scenes/simplex_sampler.py, since it seems like it's at the heart of this diff. The code needs to be more understandable before I think I'll be able to provide the review at the level I want to. I've left comments that hopefully point you in the right direction, and I've asked Slava if he can help you here as well.

I'm excited about the noise generators you're using, and the speed of map generation you've achieved, so I hope we'll be able to get this to a place where we'll be able to use it.

configs/game/map_builder/mapgen_simsam_arbitrary_tilted_lattice.yaml
```diff
@@ -0,0 +1,37 @@
...
+height: ${uniform:20,80,40}
+
+root:
+  _target_: mettagrid.map.scenes.room_grid.RoomGrid
+  rows: 3
+  columns: 3
+
+  children:
+  - scene:
+      _target_: mettagrid.map.scenes.simplex_sampler.SimplexSampler
```
berekuk, line 13: I'd recommend moving these scenes to separate files. When `children[*].scene` is a string, it's interpreted as a filename, e.g. `./configs/scenes/simplex/arbitrarily_tiled_lattice.yaml`.

Then it would be decoupled from this specific 3x3 `RoomGrid` layout.

configs/game/map_builder/mapgen_simsam_arbitrary_tilted_napkin.yaml
```diff
@@ -0,0 +1,37 @@
...
+      _target_: mettagrid.map.scenes.simplex_sampler.SimplexSampler
+      layers:
+      - fn : mettagrid.map.scenes.simplex_sampler.fn11     # arbitrary tilted lattice-like terrain
+        saturation: ${choose:1.9,2.0,2.1,2.2,2.3,2.4,2.5,3.0,4.0}
+        params:
+          lx: ${uniform:1,3}                              # scaling along x
+          ly: ${uniform:1,3}                              # scaling along y
+          t: ${uniform:-0.125,0.125}                      # tilt angle
+          ox: ${choose:2,3,4,5,6,7,8,9,10,11,12,13,14,15} # amplitude of walls along x+y
+          oy: ${choose:2,3,4,5,6,7,8,9,10,11,12,13,14,15} # amplitude of walls along x-y
```
berekuk: This can be expressed as `${int:${uniform:2,16}}`.

mettagrid/config/room/utils.py
berekuk, line 1: AFAICT this code is unused, delete?

mettagrid/map/tests/mapgen_manual_drawing.py
berekuk, line 1: With the recent upstream changes, this script should be in `tools/map/` (there's `tools.map.gen` and `tools.map.view` there now).

The combo of `python -m tools.map.gen --count 32 --output-uri ./maps` and `python -m tools.map.view ./maps/FILENAME.yaml` already covers some part of it, but I can see how a separate script for quickly generating images could also be useful.

If you want to keep it, then it should use argparse (see [this comment](https://github.com/Metta-AI/mettagrid/pull/108#discussion_r2051629381) and other `tools.map` tools) and take the config name and the count as parameters. And probably an output dir too.

mettagrid/map/scenes/simplex_sampler.py
berekuk, line 1: This code (and other code in this PR) is not formatted; we recently added Ruff for linting and formatting, so you should install it and apply `ruff format`.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+        wall_size: int = 1,
```
berekuk: `room_size` and `wall_size` look like dead code, not used

configs/game/map_builder/mapgen_simsam_arbitrary_tilted_lattice.yaml
```diff
@@ -0,0 +1,37 @@
+_target_: mettagrid.map.mapgen.MapGen
+
+width: ${uniform:20,80,40}
+height: ${uniform:20,80,40}
```
berekuk: In the current version in `main` branch, `uniform` takes two parameters. The old three-parameters sampling-aware version got renamed to `sampling`.

(I'm not sure which one we want here. The difference is that the three-parameter version by default always takes the last parameter, and adds randomness only when top-level `sampling` is not 0.)

pyproject.toml
```diff
@@ -34,6 +34,7 @@ include = [
 python = ">=3.10,<4.0"
 numpy = "^1.21.0"
 cython = "^3.0.11"
+opensimplex = "^0.4.5.1"
```
berekuk, line 37: This needs to be added to `requirements.txt` too.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,408 @@
...
+    def normalize_array(self, room: np.ndarray) -> np.ndarray:
+        norm_arr = (room - np.min(room)) / (np.max(room) - np.min(room))
+        return norm_arr
+    
+    def terrain_map(self, layer: Layer) -> np.ndarray:
+        simplex = OpenSimplex(self._rng.integers(0, 2**31-1))
+
+        xa = np.array(range(self._width))
+        ya = np.array(range(self._height))
+        terrain = np.array([simplex.noise2(*layer.fn(x, y, self._width, self._height, **layer.params)) for y in ya for x in xa]) # fn function dictates where and how fast noise will be sampled for each pixel in a room, absolute value is less important than derivative of this function
```
berekuk: numpy doesn't have any speed advantages over pure python when operations are not vectorized (actually, it's slower), so you could simply do `for y in range(self._height) for x in range(self._width)` here.

berekuk (changes requested):
I'm still figuring out your `fix_map` algorithm, will comment on it separately.

mettagrid/map/scenes/simplex_sampler.py
Dmitry273: default values sort of represent the least distorted version of a sampling function, I've added them for myself to not forget what parameters generate the "normal" result, I've added description for those reasons in now separate file with all sampling functions

mettagrid/map/scenes/terraingen.py
Dmitry273: I've made my own internal version for connectivity which fits better with my approach and works fast enough as well, but I agree that connectivity of several recursively builded maps should be done with mettagrid.map.scenes.make_connected.MakeConnected

mettagrid/config/room/utils.py
Dmitry273: I personally don't use it anymore, can't say for others

.gitignore
Dmitry273: I am using poetry

mettagrid/config/room/utils.py
berekuk: I mean the functions that you added yourself

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        # At some point it starts to climb walls since everything else is filled
+        # the terrain dictates which walls are easier to climb
+        # As soon as it touches another non-wall cell, it breaks the route that lead to it
+        # works really fast and breaks the weakest and closest to 'empty' state parts of terrain first
+        # designed to work with 2d np.arrays of ints from 0 to 255
+
+        distances = defaultdict(lambda: [float("inf"), tuple()])
+        distances[start] = [0, start]
+        front = PriorityQueue()
+        front.put((0, [start, start]))
```
berekuk, line 137: `[start, start]` is a fixed size list, so could be a tuple too.

In fact, I think the PQ could be a single tuple, `(0, start, start)`?

The comment about the semantics would be useful here, btw, both for `front` and for `distances`.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        while not front.empty():
+            energy, pair = front.get()
+            current, previous = pair
+
+            for next_dir in dir_list:
+                next_pos = (current[0] + next_dir[0], current[1] + next_dir[1])
+                if out_of_bound(next_pos[0], next_pos[1]):
+                    continue
+                if terrain[next_pos[0]][next_pos[1]] <= self.cutoff:
+                    current_energy = (
```
berekuk, line 151: This would be more readable if you modified `energy` instead of introducing the new variable.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        distances[start] = [0, start]
+        front = PriorityQueue()
+        front.put((0, [start, start]))
+        seen = set()
+        seen.add(start)
+        # store energy position and how we got there
+
+        while not front.empty():
+            energy, pair = front.get()
+            current, previous = pair
```
berekuk, line 144: `current_pos`, `previous_pos`, etc.?
Would be consistent with `next_pos` and make the algorithm more readable.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+                    current_energy = (
+                        energy + 1 + self.cutoff - terrain[next_pos[0]][next_pos[1]]
+                    )
+                else:
+                    current_energy = energy
+                    if energy > 0 and next_pos not in seen: # if we are forced to climb on walls (base energy is more than 0) and we are seeing new non-wall part, we break all the walls that got us here
+                        current_energy = 0
+                        distances[next_pos] = [current_energy, current]
+                        distances[current][0] = 0
+                        terrain[int(current[0])][int(current[1])] = 255  # type: ignore I don't know why it shows warnings there, but it works
```
berekuk, line 160: I don't think `int(...)` casting is useful here

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+
+        while not front.empty():
+            energy, pair = front.get()
+            current, previous = pair
+
+            for next_dir in dir_list:
+                next_pos = (current[0] + next_dir[0], current[1] + next_dir[1])
+                if out_of_bound(next_pos[0], next_pos[1]):
+                    continue
+                if terrain[next_pos[0]][next_pos[1]] <= self.cutoff:
```
berekuk, line 150: I find myself constantly forgetting whether low values are walls or empty spaces, can you add a comment here?

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+                        current_energy = 0
+                        distances[next_pos] = [current_energy, current]
+                        distances[current][0] = 0
+                        terrain[int(current[0])][int(current[1])] = 255  # type: ignore I don't know why it shows warnings there, but it works
+                        front.put((current_energy, [next_pos, current]))
+                        front.put((current_energy, [current, previous]))
+                        seen.add(next_pos)
+
+                        while True:
+                            if distances[previous][0] == 0:
```
berekuk, line 166: Put this in `while` condition.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+                    )
+                else:
+                    current_energy = energy
+                    if energy > 0 and next_pos not in seen: # if we are forced to climb on walls (base energy is more than 0) and we are seeing new non-wall part, we break all the walls that got us here
+                        current_energy = 0
+                        distances[next_pos] = [current_energy, current]
+                        distances[current][0] = 0
+                        terrain[int(current[0])][int(current[1])] = 255  # type: ignore I don't know why it shows warnings there, but it works
+                        front.put((current_energy, [next_pos, current]))
+                        front.put((current_energy, [current, previous]))
```
berekuk, line 162: I very strongly suspect that the previous 3-4 lines are not necessary because you repeat the same thing in the loop below. Can you find how to get rid of the duplicate code?

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+        # works really fast and breaks the weakest and closest to 'empty' state parts of terrain first
+        # designed to work with 2d np.arrays of ints from 0 to 255
+
+        distances = defaultdict(lambda: [float("inf"), tuple()])
+        distances[start] = [0, start]
+        front = PriorityQueue()
+        front.put((0, [start, start]))
+        seen = set()
+        seen.add(start)
+        # store energy position and how we got there
```
berekuk, line 140: This comment is misplaced, I think it belongs to `distances` dict?

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+                next_pos = (current[0] + next_dir[0], current[1] + next_dir[1])
+                if out_of_bound(next_pos[0], next_pos[1]):
+                    continue
+                if terrain[next_pos[0]][next_pos[1]] <= self.cutoff:
+                    current_energy = (
+                        energy + 1 + self.cutoff - terrain[next_pos[0]][next_pos[1]]
+                    )
+                else:
+                    current_energy = energy
+                    if energy > 0 and next_pos not in seen: # if we are forced to climb on walls (base energy is more than 0) and we are seeing new non-wall part, we break all the walls that got us here
```
berekuk, line 156: This is the only place where you're checking `seen`. I think it's possible to remove `seen` and check for distance here? They're kind of the same thing; you could break through the wall and get back to the part that's already connected, but then its distance would be `0`, and you could check for that.

mettagrid/map/scenes/simplex_sampler.py
```diff
@@ -0,0 +1,179 @@
...
+                            front.put((0, [previous, distances[previous][1]]))
+                            terrain[int(previous[0])][int(previous[1])] = 255  # type: ignore
+                            previous = another
+
+                if current_energy >= distances[next_pos][0]:  # type: ignore
+                    continue
+
+                distances[next_pos] = [current_energy, current]  # type: ignore
+                front.put((current_energy, [next_pos, current]))
+                seen.add(next_pos)
```
berekuk, line 179: General thoughts on this algorithm:

1) The idea is very good, I like it.

2) The implementation could be clearer, see comments above.

---

3) I've been thinking how to adapt it to `MakeConnected` (which has several issues with it).

I'm pretty sure that the integer terrain could be adapted somehow (and could even be useful - right now `MakeConnected` treats walls and previously placed objects as having the same weight, and it could be good to prioritize one or another for removal).

But another concern is that `MakeConnected` has randomness, and that's a nice thing to have.

If want to compare them for yourself, you can check out `configs/game/map_builder/mapgen_bsp_plus_connected.yaml` as an example, or make your own config that calls `RoomGrid` with a very fine grid of isolated 1x1 cells, and then call `MakeConnected` on top of it. `MakeConnected` works decently on top of BSP; it deletes almost everything from 1x1 `RoomGrid` because it connects everything to a single area.

OTOH, your algorithm removes the minimal amount of walls, but it turns 1x1 `RoomGrid` into a boring comb-like shape (I tried it - copy-pasted your code and converted the grid to the terrain and then back). The perfect approach, I think, would produce something very maze-like. Maybe shuffling the directions in your code would be enough, but I'm not sure.

Anyway, this part is not necessary for this PR to go in, just something to keep in mind for the future.

mettagrid/map/scenes/simplex_sampler.py
Dmitry273: I renamed them to be more readable, I think you can use one variable, but it will make it less intuitive

mettagrid/map/scenes/simplex_sampler.py
Dmitry273: yes, you're right, I was trying to get rid of warnings, but it didn't helped anyway and I forgot to remove it after I've just suppresed warnings

PR #80, opened 2025-04-14
fix: refactor resolvers

rjwalters:
This PR improves code organization and maintainability by:

1. Moving OmegaConf resolvers from `mettagrid_env.py` into a dedicated `resolvers.py` module
2. Simplifying the resolver logic, particularly for the `uniform` resolver
3. Expressing default values (when sampling = 0) directly in the environment config files using the `if` resolver
4. Bringing codebase into compliance with ruff linter standards
   - Standardizing import order
   - Fixing line length issues
   - Improving formatting consistency
   - Removing unused imports

The refactoring maintains backward compatibility with existing configurations while making the system more modular and easier to extend with new resolvers in the future.

Key changes:
- Created new `resolvers.py` with all OmegaConf resolvers
- Added re-export in `__init__.py` for simplified imports
- Updated maze configuration files to use the new resolver pattern
- Fixed linting issues across multiple files


.vscode/settings.json
```diff
@@ -1,3 +1,35 @@
 {
-    "python.analysis.typeCheckingMode": "basic"
-}
+    // ruff format on save is configured in metta.code-workspace
```
sasmith, line 2: Not valid json. If vscode is okay with it, I am too.

mettagrid/resolvers.py
```diff
@@ -0,0 +1,73 @@
...
+    return a * b
+
+
+def oc_add(a: float, b: float) -> float:
+    return a + b
+
+
+def oc_make_odd(a: float) -> int:
+    """
+    Ensure a value is odd and at least 3.
```
sasmith, line 38: Weird function.

sasmith (changes requested):
Can you split this diff? I'd like to be able to review the innocuous lint-only changes (whitespace, import reordering, other such trivial stuff) from less obviously innocuous stuff like changing the `uniform` calls.

This is a request I'll try to make pretty much any time a PR contains "more than one big idea" or mixes things I should rubber stamp with things I need to think about.

.vscode/settings.json
rjwalters: vscode uses a relaxed JSON parser that supports js style comments 

we could use JSON style comments:
```json
{
  "_comment": "ruff format on ..."
}
```

but this seems worse.

PR #82, opened 2025-04-14
Allow infinite cooldown

daphnedemekas:
If cooldown is -1, converter never resets


mettagrid/objects/converter.hpp
```diff
@@ -113,10 +113,15 @@ class Converter : public HasInventory {
             // Start cooldown phase
             this->cooling_down = true;
             this->event_manager->schedule_event(Events::CoolDown, this->cooldown, this->id, 0);
-        } else {
+        }
```
sasmith, line 116: 116 and 117 should be on a single line, but linting should fix this.

sasmith (approved):

PR #84, opened 2025-04-15
feat: refactor resolvers

rjwalters:
# Refactor OmegaConf resolvers

This PR reorganizes our configuration resolvers to improve maintainability and feature coverage. The primary changes include:

- Created a dedicated `resolvers.py` module to centralize all OmegaConf resolver definitions
- Renamed `uniform` resolver to `sampling`
- Added several new utility resolvers for common operations (if, equals, add, multiply, etc.)
- Updated configuration files to use the new resolver system
- Fixed indentation and formatting inconsistencies in YAML configuration files

## Implementation Details

The renamed `sampling` resolver now asserts sampling is < 1:

1. Uses the global config `sampling` parameter (0.0-1.0) to control environment randomness
3. When `sampling=0`, exact center values are used (deterministic)
4. When `0<sampling<=1`, values are randomly selected within a scaled range around the center
5. When `sampling > 1` we will throw an error

This approach clarifies how we are configuring environments while maintaining backward compatibility.


mettagrid/resolvers.py
```diff
@@ -0,0 +1,118 @@
...
+
+def oc_make_integer(value: float) -> int:
+    return int(round(value))
+
+
+def oc_equals(a, b) -> bool:
+    return a == b
+
+
+def oc_scaled_range(lower_limit, upper_limit, center, *, root=None):
```
sasmith: Thank you for the renaming. I think these could be typed too?

sasmith (approved):

PR #90, opened 2025-04-16
feat: Random Seed Management and Uniform List Resolver

rjwalters:
# Random Seed Management and Uniform List Resolver

This PR introduces two enhancements to our OmegaConf resolvers:
1. Improved random seed management with a time-based default seed
2. New `uniform_list` resolver for generating multiple random values efficiently

## Random Seed Handling
- **Default Behavior**: When no seed is specified, our module defers to NumPy's built-in entropy sources
- **Explicit Seeds**: For reproducibility, you can still set explicit seeds when needed

#### Dynamic (Default) Randomness
```yaml
# Config uses different random values on each run
width: ${uniform:10,50}
height: ${uniform:20,100}
```

#### Setting a Specific Seed for Reproducibility
```yaml
# First set a specific seed
_: ${set_random_seed:42}

# All subsequent random operations will be consistent across runs
width: ${uniform:10,50}  # Always returns the same value when seed is 42
height: ${uniform:20,100}  # Always returns the same value when seed is 42
```

## Uniform List Resolver

The new `uniform_list` resolver generates multiple random values in a single efficient operation.

### Syntax
```
${uniform_list:<min>,<max>,<count>}
```

### Examples

```yaml
# Generate a list of 5 random values between 0 and 1
points: ${uniform_list:0,1,5}

# Generate 10 random integers (will be rounded)
coords: ${int:${uniform_list:1,100,10}}

# Practical application: random grid of coordinates
grid:
  x_coords: ${uniform_list:0,100,10}
  y_coords: ${uniform_list:0,100,10}
```

## Benefits

- **Better Defaults**: Unique sequences by default, reproducible when needed
- **Improved Performance**: `uniform_list` is more efficient than multiple calls
- **Flexibility**: Both global and per-operation seed control
- **Consistency**: All random operations follow the same seeding rules

### DIFF
--- a/mettagrid/resolvers.py
+++ b/mettagrid/resolvers.py
@@ -12,7 +12,9 @@ def oc_if(condition: bool, true_value: T, false_value: T) -> T:
     return true_value if condition else false_value
 
 
-def oc_uniform(min_val: Numeric, max_val: Numeric) -> float:
+def oc_uniform(min_val: Numeric, max_val: Numeric, seed: Optional[int] = None) -> float:
+    if seed is not None:
+        np.random.seed(seed)
     return float(np.random.uniform(min_val, max_val))
 
 
--- a/mettagrid/resolvers.py
+++ b/mettagrid/resolvers.py
@@ -18,6 +18,13 @@ def oc_uniform(min_val: Numeric, max_val: Numeric, seed: Optional[int] = None) -
     return float(np.random.uniform(min_val, max_val))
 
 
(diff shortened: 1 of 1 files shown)


rjwalters:
I think this approach is still flawed... we are resolving the config (i.e. selecting random values) like this:

```python
    def _resolve_original_cfg(self):
        """Create a new resolved environment configuration from the template."""
        # this is where values will be drawn from a distribution
        env_cfg = OmegaConf.create(copy.deepcopy(self._original_cfg))
        OmegaConf.resolve(env_cfg)
        return env_cfg
```

in `mettagrid_env.py`

so it isn't clear how we would avoid resetting the random seed each time

PR #94, opened 2025-04-16
feat: enforce Ruff in CI check

rjwalters:
This PR adds a lint check on PRs that is tied to the configuration at https://github.com/Metta-AI/metta/blob/main/ruff.toml so that we have only one source of truth.


rjwalters:
This PR is not going to pass CI and is separated out here for review convenience. We should close this and just merge in #95

rjwalters:
I accidentally merged #95 into this branch instead of master... oops

PR #95, opened 2025-04-16
fix: bring repo into compliance with Ruff linter

rjwalters:
Assuming we're happy with the CI check in #94, this PR corrects all outstanding lint issues reported so that CI can pass.


configs/maze.yaml
```diff
@@ -1,7 +1,7 @@
 defaults:
   - mettagrid
-  - _self_
   - game/map_builder: maze
+  - _self_
```
berekuk, line 4: oops, right, I'm not sure why I put them in this order (not that it matters)

mettagrid/map/scene.py
```diff
@@ -15,8 +16,8 @@ class TypedChild(TypedDict):
 
 # Base class for all map scenes.
 class Scene:
-    def __init__(self, children: list[TypedChild] = []):
-        self._children = children
+    def __init__(self, children: Optional[List[TypedChild]] = None):
+        self._children = children if children is not None else []
```
berekuk, line 20: `children or []` would be shorter / more readable (same for other similar places)

.github/workflows/build.yml
```diff
@@ -100,4 +100,4 @@ jobs:
         run: |
           source venv/bin/activate
           pytest tests/test_basic.py
```
berekuk, line 102: unrelated to Ruff, but: there are more tests my PR added `tests/test_pattern.py`.

How about `pytest tests`?

mettagrid/map/node.py
```diff
@@ -57,15 +57,21 @@ def select_areas(self, query) -> list[Area]:
...
+                            match = True
+                            for tag in tags:
+                                if tag not in area.tags:
+                                    match = False
+                                    break
+                            if match:
+                                selected_areas.append(area)
+                else:
+                    # Handle the case where "where" doesn't have expected structure
+                    selected_areas = areas
```
berekuk, line 74: Maybe `raise`?
(unsure if matters if we move to structured configs)

PR #96, opened 2025-04-16
Rename actions -> metta_action_handler

sasmith:
Renamed `actions.pyx/pxd` to `metta_action_handler.pyx/pxd` and updated all imports across the codebase, to make the filename match the class it defines. This is in service of cleaning up and removing the cython we have in our codebase.

Tested by compiling and doing basic training.

PR #92, opened 2025-04-16
rename action -> action_handler

sasmith:
This is mostly a file rename, since having the files named differently than their main class is a pain. This also removes `Action` which is unused.

Directionally, this is towards getting rid of cython code.

Tested by compiling, and doing basic local training.

PR #98, opened 2025-04-17
Move rewards from grid_env to agents

sasmith:
This moves reward -- at least the single-step accumulation of reward -- to agents. This move reduces the dependency of actions on grid_env. Once that dependency is gone, we can convert actions out of cython.

This also mostly feels cleaner.

Test:
Ran some benchmark perf tests to make sure something wasn't obviously off about this: `python deps/mettagrid/tests/benchmark_env_perf.py`

### DIFF
--- a/mettagrid/actions/attack.pyx
+++ b/mettagrid/actions/attack.pyx
@@ -35,7 +35,7 @@ cdef class Attack(MettaActionHandler):
         if actor.inventory[InventoryItem.laser] == 0:
             return False
 
-        actor.update_inventory(InventoryItem.laser, -1, &self.env._rewards[actor_id])
+        actor.update_inventory(InventoryItem.laser, -1)
 
         cdef short distance = 0
         cdef short offset = 0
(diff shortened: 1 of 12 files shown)


sasmith:
* **#99** <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/99?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#98** <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/98?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/98?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#96** <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/96?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#92** <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/92?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

daveey (changes requested):
Agent should take a ptr to env._rewards rather than allocating it's own reward memory. that would allow us to use the rewards[:] = 0, and not require copying rewards over every time step.

sasmith:
Okay. I considered this initially, but decided it was basically the same in the end. I.e., this version means we need to deal with the indirection at the end of the step. If the agent takes a pointer, then we need to deal with indirection any time agent's rewards are updated.

But will change.

PR #100, opened 2025-04-17
fix: address comments on PR #95

rjwalters:
PR #95 was merged a bit early before all comments were addressed

### **Code Style**
- [x] Replace explicit `x if x is not None else []` with more concise `x or []` where safe and appropriate.
  - Example:  
    ```python
    self._children = children if children is not None else []
    ```
    becomes:
    ```python
    self._children = children or []
    ```
  - Location: `mettagrid/map/scene.py` and any similar patterns in other files

### **CI Workflow**
- [x] Update test runner in `.github/workflows/build.yml`:
  - Replace:
    ```bash
    pytest tests/test_basic.py
    ```
  - With:
    ```bash
    pytest tests
    ```
  - Rationale: Ensure all tests (e.g., `tests/test_pattern.py`) are included

### **Logic/Robustness**
- [x] In `mettagrid/map/node.py`, consider raising an error (or logging) when `where` does not match expected structure.
  - Current fallback:  
    ```python
    selected_areas = areas
    ```
  - Reviewer suggestion: `raise` may be more appropriate in some contexts

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,5 +99,5 @@ jobs:
           PYTHONPATH: ${{ github.workspace }}
         run: |
           source venv/bin/activate
-          pytest tests/test_basic.py
+          pytest tests
           python tests/test_kinship.py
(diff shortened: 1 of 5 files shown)


berekuk:
lgtm, thanks!

PR #102, opened 2025-04-17
MapGen: Scene files; RandomScene scene; more random samplers; clearer Auto scene syntax

berekuk:
## scene files
Any scenes can now be loaded by config reference, `scene: /scenes/wfc/blob`.

Right now this works through Hydra but I expect I'll change it to use OmegaConf directly soon, because:
- unwrapping the config ("config_from_path" from main metta repo, I used some parts of it for this) is weird
- I don't think defaults are useful here
- I expect that plain OmegaConf configs will be easier to share between mettagrid and metta; Hydra conventions are too strict for this use case

## RandomScene
Should be self-explanatory:
```
  _target_: mettagrid.map.scenes.random_scene.RandomScene

  candidates:
    - scene: /scenes/convchain/dungeon
      weight: 2
    - scene: /scenes/convchain/blob
      weight: 1
```

## more random samplers

In `mettagrid.map.utils.random`. I'm keeping an eye on the possibility to replace these with resolvers, but for now I still think that arrays are less risky here.

(I have a plan for fully reproducible scenes that will require a careful seed/rng management, and that could be hard or impossible to do with resolvers. This idea is still in design stage though, doing some experiments to see what works best.)

## clearer Auto scene syntax

Self-explanatory, see `mapgen_auto.yaml`.

---

I expect to do a bunch of small changes like these in the next few days. Not sure where to put the boundary where to send the PR, but sending these now so that we don't end with another 3000 lines diff...


sasmith (approved):
> Right now this works through Hydra but I expect I'll change it to use OmegaConf directly soon

My familiarity with Hydra is all new, and my familiarity with OmegaConf is only through Hydra. So I'll be happy to have this discussion when it shows up.

PR #103, opened 2025-04-17
Prioritized environment replay

daphnedemekas:
I asked Claude to replace my MettaGridEnvSet with a prioritized environment selector

Looks like its selecting environments based on reward, so lower reward means higher priority. 

Didn't work hard on this, I know that @rjwalters is thinking about the same thing and also potentially Joseph? 

Might try this out on some navigation stuff and see if it helps. But don't feel strong urgency on merging


rjwalters (commented):
I think we might want to keep both strategies for backwards compatibility. This could also be a good time to split the multi env into a separate file

rjwalters:
I like this idea -- let me have a couple of days to work up a version

PR #108, opened 2025-04-20
S3 maps generation and loading

berekuk:
Plus documentation - check out https://github.com/Metta-AI/mettagrid/blob/78921f7856b72d740dcf76322232c3fe6fbb08bc/docs/mapgen.md


mettagrid/objects/constants.pyx
```diff
@@ -0,0 +1,7 @@
+from mettagrid.objects.constants cimport ObjectTypeNames, ObjectTypeAscii
+
+def get_object_type_names():
+    return ObjectTypeNames
+
+def get_object_type_ascii():
+    return ObjectTypeAscii
```
berekuk: I use ascii symbols from Cython `render_ascii` code, so I needed these to decode the symbols back to names.

This is not 100% safe, btw - in ascii, agent types are lost. Would be good to clean up later.

tests/mapgen.py
berekuk, line 1: Moved to `tools/`, as requested by David in Asana

configs/index_s3_maps.yaml
```diff
@@ -0,0 +1,11 @@
+defaults:
```
daveey: do we need to index, can we just read the dir content from s3?

tools/mapgen.py
```diff
@@ -0,0 +1,69 @@
...
+
+    elif mode == "none":
+        pass
+
+    else:
+        raise ValueError(f"Invalid show mode: {mode}")
+
+
+@hydra.main(version_base=None, config_path="../configs", config_name="mapgen")
+def main(cfg):
```
daveey: can we use argparse here isntead of hydra? i think it's a better match for utilities. we can still use hydra.compose() to load the map config, but other than that, i think this adds unnecessary configs and complexity.

python -m tools.mapgen --output-uri=s3://softmax-public/maps --max-maps=5 --num-workers=4 --env=env/mettagrid/simple 

it would loop while output-uri has fewer than max-maps, creating and dumping a map in there with a random name. it would use a multiprocessing pool to do this in parallel with --num-workers

i don't want to maintain an index file, because it makes paralell generation harder, and listing a directory is easy

tools/mapgen.py
```diff
@@ -0,0 +1,69 @@
...
+        if target_name is None:
+            random_suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
+            target_name = f"map_{random_suffix}.yaml"
+
+        target_uri = os.path.join(cfg.mapgen.target.dir, target_name)
+        ascii_map = StorableMap.from_env(env, gen_time=gen_time)
+        ascii_map.save(target_uri)
+
+    # Show the map if requested
+    show_env(env, cfg.mapgen.show)
```
daveey: you can add --view flag that shows each map. by default output-uri can be /tmp/metta-maps and max-maps can be inf, so that when view is used, it dumps a new random map into /tmp/ and displays it

mettagrid/objects/constants.pyx
daveey: we should clean this up. i used ascii in the env for fast debugging, and it was never meant to be used for anything else. instead of getting ascii from the env, can you add a "symbols" config somewhere, and use that instead?

mettagrid/map/utils/serialization.py
```diff
@@ -0,0 +1,154 @@
...
+        return content
+
+    def width(self) -> int:
+        return len(self.lines[0])
+
+    def height(self) -> int:
+        return len(self.lines)
+
+    @staticmethod
+    def from_env(env: MettaGridEnv, gen_time: float) -> "StorableMap":
```
daveey: this seems un-nessary. we should not need an env to get a map, since maps are needed to make an env.

mettagrid/map/scenes/remove_agents.py
```diff
@@ -0,0 +1,25 @@
+from mettagrid.map.node import Node
+from mettagrid.map.scene import Scene, TypedChild
+
+
+class RemoveAgents(Scene):
```
daveey, line 5: i don't love this pattern. i don't think i understand the problem you're trying to solve. we don't have to define game.num_agents, but we do need to compute it from the map, and make sure it doesn't change during a training run. open to ideas on how to do that best.

mettagrid/map/scenes/nop.py
```diff
@@ -0,0 +1,14 @@
+from mettagrid.map.node import Node
```
daveey, line 1: do we need this if we don't removeAgents?

mettagrid/map/load_random.py
```diff
@@ -0,0 +1,31 @@
...
+    Load a random map from a list of pregenerated maps.
+
+    The directory must contain a file `index.txt` that lists all the maps in the
+    directory. (Listing S3 objects would be too slow because of pagination.)
+
+    The index file can be produced with the following command:
+        python -m tools.index_s3_maps index_s3_maps.dir=s3://...
+    """
+
+    def __init__(self, index_uri: str, extra_root: SceneCfg | None = None):
```
daveey: just support list-dir

daveey (dismissed):

configs/index_s3_maps.yaml
berekuk: It might be somewhat slow with 10k+ maps - s3 only allows to list 1k maps at a time, so would require pagination.

`aws s3 ls s3://softmax-public/maps/` for me takes 1 second. might be because I'm far from the US or because of https cold start, but I didn't want to take a risk and waiting for 10 seconds for each map. Maybe I should measure it and pagination is well-optimized?

Even keeping the index on s3 is somewhat slow and could be 10x multiplier on data sent... Which is why I tried to keep that part agnostic.

I could add lazy indexing & local cache if you think it'd be better (but then I'd need to come up with conventions on where to cache, which is another layer of configuration, so I wanted to avoid that).

mettagrid/map/scenes/remove_agents.py
berekuk: I'm not happy about `game.num_agents` config duplication either (previously discussed [here](https://github.com/Metta-AI/mettagrid/pull/73#issuecomment-2791111162)), but I think this handles a valid scenario even if we get rid of `num_agents`.

If we have the pre-generated maps, it's reasonable to want to run modifications of them without regenerating the map from scratch. Compare: in Starcraft, you can start the same for N players with M<N players.

Scenes already support drawing on top of existing maps, so this is not that much of a hack. (My initial take was to call this `ForceAgentCount` and support both increasing and decreasing the number of agents, but sharing code with `Random` scene was too annoying and I used two scene calls instead.)

In general, my policy for adding new scenes is to add things that might be useful, and we can remove the stuff that's not used later.

mettagrid/map/scenes/nop.py
berekuk: Technically we don't need Nop even with RemoveAgents, it's here just for a bit more clear config.

Load random uses this:

```
extra_root:
  _target_: mettagrid.map.scenes.nop.Nop

  children:
    - where: full
      scene:
        _target_: mettagrid.map.scenes.remove_agents.RemoveAgents

    - where: full
      scene:
        _target_: mettagrid.map.scenes.random.Random
        agents: 40
```

Instead, it could be:
```
extra_root:
  _target_: mettagrid.map.scenes.remove_agents.RemoveAgents

  children:
    - where: full
      scene:
        _target_: mettagrid.map.scenes.random.Random
        agents: 40
```

These are equivalent, I think the former reads a bit better, but not too stuck up on it.

(I'm not sure `extra_root` is the best option here, maybe `Load` and `LoadRandom` generators should accept the plain list of scenes, or maybe even `MapGen` generator should take the list of scenes instead of a single root scene. `where: full` is kind of ugly.)

berekuk:
- rewrote `mapgen.py` and `index_s3_maps.py` with argparse
- changed `LoadRandom` to list s3 directly; moved previous version with index file to `LoadRandomFromIndex` in case we need it later
- changed `StorableMap` so that it works from numpy grid instead of env's `render_ascii` (not all symbols are supported; I'll upgrade the format for storing maps soon, want to avoid pickle, I guess the easiest option is CSV? I don't love that it's less readable though)
- remove a few places where Hydra was used; scenes in mapgen configs are now addressed by the path (unfortunately this means that scene paths are absolute from the mettagrid root... will continue to think how to improve, ideally we want to track DictConfig source somehow, wrap `OmegaConf.load` so that it returns our patched version of DictConfig that's file source-aware?)

mettagrid/map/node.py
```diff
@@ -58,9 +59,9 @@ def select_areas(self, query) -> list[Area]:
                 selected_areas = [self._full_area]
             else:
                 # Type check and handling
-                if isinstance(where, dict) and "tags" in where:
+                if (isinstance(where, dict) or isinstance(where, DictConfig)) and "tags" in where:
```
sasmith: (a) This smells funny, and makes me want to move in the direction of having a "where" object. I don't think we need to do that yet.
(b) `isinstance`'s second argument can be a tuple of classes. So this could be `if isinstance(where, (dict, DictConfig)): ...`

mettagrid/map/utils/s3utils.py
```diff
@@ -0,0 +1,66 @@
...
+        raise ValueError(f"URI {uri} is not an S3 URI")
+
+    s3_parts = uri[5:].split("/", 1)
+    bucket = s3_parts[0]
+    key = s3_parts[1]
+    return bucket, key
+
+
+def get_s3_client():
+    import boto3
```
sasmith, line 18: It's worth a comment on why you're dynamically importing boto3. I assume this is to allow us to not crash if boto3 isn't installed, unless people try to use s3?

mettagrid/map/utils/s3utils.py
```diff
@@ -0,0 +1,66 @@
...
+def is_s3_uri(uri: str) -> bool:
+    return uri.startswith("s3://")
+
+
+def parse_s3_uri(uri: str) -> tuple[str, str]:
+    if not uri.startswith("s3://"):
+        raise ValueError(f"URI {uri} is not an S3 URI")
+
+    s3_parts = uri[5:].split("/", 1)
+    bucket = s3_parts[0]
```
sasmith: This is fine, but consider `bucket, key, *rest = s3_parts`

mettagrid/map/utils/s3utils.py
```diff
@@ -0,0 +1,66 @@
...
+    return session.client("s3")
+
+
+def save_to_uri(text: str, uri: str):
+    if is_s3_uri(uri):
+        bucket, key = parse_s3_uri(uri)
+        s3 = get_s3_client()
+        s3.put_object(Bucket=bucket, Key=key, Body=text)
+    else:
+        with open(uri, "w") as f:
```
sasmith: Unless I'm somehow really confused, this won't open `uri` as a uri, but as a file. For example, you should be passing `path/my_file.txt` rather than `file://path/my_file.txt`.

My preference would be for us to be using `uri`s, vs mixing s3 uris and filenames, so I think this should be changed to expect and strip the `file://` prefix.

configs/index_s3_maps.yaml
```diff
@@ -0,0 +1,7 @@
+index_s3_maps:
+  # S3 directory, `s3://.../dir/`
```
sasmith: I think this is misleading, since `dir` should look like `s3://bucket/directory/`, but the usage of `dir` in the comment makes look like `dir` should just be `directory`.

For clarity (and assuming we want to keep `dir`'s _usage_ unchanged), I think `dir` should be renamed to something like `uri_prefix`. Notably, this parameter is expected to contain the scheme and the path, where as a "dir" would not include the scheme.

configs/index_s3_maps.yaml
```diff
@@ -0,0 +1,7 @@
...
+  dir: ???
+  # By default the index will be saved on S3 to `dir/index.txt`.
```
sasmith: I think this comment should be
```
# The save location for the map index. Defaults to `${s3_prefix}/index.txt`.
# If provided, this should be an s3 or file URI.
```

(For this to be true, we need to change this to want a file _uri_ vs just a filename.)

I think this should also be renamed to something clearer that `target`, since the index is a small part of what we're doing here. Maybe `index_target_uri`?

docs/mapgen.md
```diff
@@ -0,0 +1,43 @@
+# Map generation
+
+## S3 maps
+
+To produce maps in bulk and store them in S3, use the following commands:
+
+### Creating maps
+
+```bash
+python -m tools.mapgen game/map_builder=mapgen_auto mapgen=s3 mapgen.s3_dir=s3://BUCKET/DIR
```
sasmith: For now, `s3_dir` is incorrect, and this should be `dir`. But as per other comments, I think we want to change this.

tools/mapgen.py
```diff
@@ -0,0 +1,139 @@
...
+            return output_uri
+
+        if not output_dir:
+            return None  # the map won't be saved
+
+        random_suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=8))
+        return f"{output_dir}/map_{random_suffix}.yaml"
+
+    if output_uri:
+        output_dir = os.path.dirname(output_uri)
```
sasmith: As in other comments, this is confusing, since I would definitely expect a uri to have a scheme, and I would expect a dir to not have a scheme.

sasmith (changes requested):

mettagrid/map/utils/s3utils.py
berekuk: I'm treating `uri` here in a loose sense, "the parameter that can be one of these syntax versions".

I tried to cover some subset of other config options in metta that are called "uri" - `policy_uri` and `eval_db_uri`. AFAICT both of these support plain file names.

I've changed to code to strip `file://` but I'd prefer to keep the support for filenames. I like running mapgen with `python -m tools.mapgen --output-dir=.`, and there'd be no auto-completion with `file://` version.

---

Btw, there's a whole rabbit hole with "how relative paths are treated in file URIs" - `file://path/to/file` is supposed to be the absolute `/path/to/file`, and none of uri functions in metta repo do this... My code does things in the same incorrect way, for now (blindly strips `file://`).

Maybe we'll be able to extract these common functions under `mettautils` after we get monorepo?

configs/index_s3_maps.yaml
berekuk: Oops, I forgot to remove Hydra configs. `index_s3_maps.py` and `mapgen.py` use argparse now!

Also forgot to update docs for argparse... fixed that too.

mettagrid/map/utils/s3utils.py
berekuk: This is an accidental feature because I refactored this code out from mapgen.py script, and I did dynamic import there for the miniscule boost in performance.

But now I notice that there's no dependency on boto3 in mettagrid, I was using metta venv, and maybe dynamic import is a good thing. So, commented.

mettagrid/map/utils/s3utils.py
sasmith: > I've changed to code to strip file:// but I'd prefer to keep the support for filenames. I like running mapgen with python -m tools.mapgen --output-dir=., and there'd be no auto-completion with file:// version.

Okay, this sounds good. I still feel squeamish about the mixing of bonafide uris and not-quite-uris, but don't need to be dogmatic.

> Maybe we'll be able to extract these common functions under mettautils after we get monorepo?

Agreed.

docs/mapgen.md
```diff
@@ -0,0 +1,43 @@
...
+### Loading maps
+
+You can load a random map from an S3 directory by using `mettagrid.map.load_random.LoadRandom` as a map builder.
+
+Check out `configs/game/map_builder/load_random.yaml` for an example config and for how to tune the number of agents in the map.
+
+Preview a random map:
+
+```bash
+python -m tools.mapgen ./configs/game/map_builder/load_random.yaml --overrides='dir=s3://BUCKET/DIR'
```
sasmith: This is confusing, since nothing in this command would make me expect I'm going to view a map.

For now, let's separate mapgen commands from view commands, since it's easier to reason about.

mettagrid/map/utils/storable_map.py
```diff
@@ -0,0 +1,128 @@
...
+import numpy as np
+from omegaconf import DictConfig, OmegaConf
+
+from mettagrid.map.mapgen import MapGrid
+
+from . import storage
+
+logger = logging.getLogger(__name__)
+
+ascii_symbols = {
```
sasmith, line 13: Okay for this diff, but let's fast-follow with consolidating these into mettagrid.yaml. E.g.,

```
  objects:
    altar:
      hp: 30
      input_battery: 3
      output_heart: 1
      max_output: 5
      conversion_ticks: 1
      cooldown: ${sampling:1, 20, 10}
      initial_items: 1
      ascii_symbol: a
...
```

tools/mapgen.py
```diff
@@ -0,0 +1,139 @@
...
+# Aggressively exit on ctrl+c
+signal.signal(signal.SIGINT, lambda sig, frame: os._exit(0))
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+ShowMode = Literal["raylib", "ascii", "ascii_border"]
+
+
+def show_map(storable_map: StorableMap, mode: ShowMode | None):
```
sasmith: As per the comment on the readme, can we split this into a separate entry point (mapview.py / mapshow / whatever). Maybe with a subdir, so we get

`python -m tools.map.gen` and `python -m tools.map.view`.

tools/mapgen.py
```diff
@@ -0,0 +1,139 @@
...
+            "timestamp": datetime.now().isoformat(),
+        },
+        config=cfg,
+    )
+    return storable_map
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--output-dir", type=str, help="Output directory, e.g. ./maps or s3://.../dir")
```
sasmith: Can we figure out how to reasonably collapse `--output-dir` and `--output-uri`? Sorry for the back and forth, and for at least some level of inconsistency on my part. Dave and I chatted, and where we've landed is

1) no output-dir, just `output-uri`.
2) `output-uri` is optional. If not provided, then assert `count` is 1 and dump to stdout.
3) If `output-uri` has a scheme, it should be `s3` or `file`. If it's not provided, assume it's `file`.
4) If `output-uri` ends with a file suffix (let's say `.` and <= 4 characters? Seems like `.yaml` matters the most, and `.txt` / `.map` / similar also seem reasonable), then assume it's the full target, assert `count` is 1, and write the output there.
5) Otherwise, assume `output-uri` is a directory (/ prefix)
`--output-uri` can be

mettagrid/map/scenes/remove_agents.py
sasmith: Okay, let's ship this for now. Wanting to update pre-existing maps makes sense. Depending on how much that happens we may want a generic set of "update map" utilities, but this seems like a good tool to figure out what we'll want.

docs/mapgen.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+### Indexing maps
+
+Optionally, you can index your maps to make loading them faster.
```
sasmith, line 49: Since this is optional, I'd like to make sure we're arming people with enough info to know if that need to learn about this (also, I know Dave is hoping we can just not worry about this). How about "This is intended to speed up reading from S3. It shouldn't change any functionality, and you should skip playing with this unless you find map loading from S3 is slow".

sasmith (changes requested):

mettagrid/map/utils/storable_map.py
berekuk: Several quick thoughts:
- ascii is not a good format to store maps anyway (see all the exceptions in `grid_object_to_ascii` function in this file; no support for colored mines and agents)
- DCSS maps have metadata on symbols in maps, `SUBST: 1=agent.prey, 2=agent.predator`, that might be interesting to try
- or just dump the full CSV with long names? sacrifice readability, but more robust
- generally, I really want to unify all our serialization formats and/or have tools for converting between them, I know we have at least four by now (this ascii, another completely different ascii in `mettagrid.config.room.ascii`, full names, int type ids...) - this is something I'm keeping in mind as high priority

tools/mapgen.py
berekuk: Done; `tools.map.gen` can still do everything it could before, because "gen-and-view" is convenient. `tools.map.view` supports the same heuristics for dir vs file and reuses "load random" code for dirs.

tools/mapgen.py
berekuk: I feel like these heuristics are going to backfire for someone eventually, but ok :) Changed according to your spec. (I don't have any better ideas...)

Note: I realize that `map.gen.view` could be a bit more clever and test "is this a dir?" empirically instead of relying on heuristics. I did it with heuristics for now (same `uri_is_file` function in both scripts), because all combinations of file/dir and local/s3 would be annoying to handle; on s3 I think it's possible that name is *both* a dir and a file, because dirs don't really exist. And also because I expect that if heuristics will misfire then `tools.map.gen` will break too.

berekuk:
Addressed everything + added CLI tests.

mettagrid/map/utils/storable_map.py
sasmith: Oh, yeah. We're with you on not using ascii in the long run, and were considering asking you to rip it out -- but it's still being used by folks.

sasmith (approved):
Approving and merging, and I'll also bring this to the metta repo.

PR #111, opened 2025-04-21
feat: add benchmarks to CI

rjwalters:
Goal is to use:

- pytest-benchmark 
- bencher github action 
- RunsOn (so that we can run tests on gpu machines on amazon)

I am also updating and repairing the test suite as I go


rjwalters:
moved to https://github.com/Metta-AI/metta/pull/216

PR #112, opened 2025-04-21
Switch MettaScope to use WebGPU.

treeform:
* Use WebGPU for faster drawing.
* Add a play button.
* Add inventory above the agents.
* Add rewards bellow the agents.
* Add which actions agents are currently doing.

image


player/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "webgpu-experiments",
```
daveey, line 2: this seems wrong

player/src/drawer.ts
```diff
@@ -0,0 +1,932 @@
...
+  // Set up the index buffer pattern once
+  setupIndexPattern(): void {
+    // For each quad: triangles are formed by indices
+    // 0-1-2 (top-left, bottom-left, top-right)
+    // 2-1-3 (top-right, bottom-left, bottom-right)
+    for (let i = 0; i < this.maxQuads; i++) {
+      const baseVertex = i * 4;
+      const baseIndex = i * 6;
+
+      this.indexData[baseIndex + 0] = baseVertex + 0; // Top-left
```
daveey, line 56: can this just be a loop over [0,1,2,2,1,3]

player/src/drawer.ts
```diff
@@ -0,0 +1,932 @@
...
+    if (this.currentQuad >= this.maxQuads) {
+      throw new Error("Max quads reached");
+    }
+
+    // Calculate base offset for this quad in the vertex data array
+    const baseVertex = this.currentVertex;
+    const baseOffset = baseVertex * 8; // Each vertex has 8 floats
+
+    // Top-left vertex
+    this.vertexData[baseOffset + 0] = topLeft.x();
```
daveey, line 129: can this be a loop over an array?

daveey (approved):
Looks good. You'll need to port it over to the new mono-repo but should just be a simple copy.

I don't have any experience with web game dev, so don't have anything meaningful to say about the logic. made a couple of minor code style suggestions.

player/src/drawer.ts
treeform: The for loop does save some lines, but it makes it harder to read, but I will make the change as requested.

player/src/drawer.ts
treeform: The for loop does save some lines, but it makes it harder to read, but I will make the change as requested.

PR #115, opened 2025-04-22
Convert actions from pyx to cpp

sasmith:
This converts action_handlers and its subclasses from pyx to cpp. This makes a few decisions:
* This converts from Python classes to cpp classes. We could have preserved there being compiled Python classes to wrap the cpp classes, but it's not clear where we would actually want to interact with the action classes in Python. However, `__init__` functions are Python functions, so `GridEnv.__init__` expect Python objects. This means we can't pass `ActionHandlers` to `__init__` directly; and instead we have an `action_handler_init` cpp function that `MettaEnv` calls.
* We configure each action with an `ActionConfig`, even though none of the actions use this. This is just preserving the config-passing we were doing before this change. I don't have a strong opinion here.
* I guess that's it. It felt like more, since the get-Python-to-deal-with-cpp-objects took a very long time for me to resolve. The error messages were very rarely helpful to me.


sasmith:
* **#115** <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Metta-AI/mettagrid/115?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

daveey (approved):

daveey:
Thanks for doing this. I'm not reviewing the details, but the overall change looks great

sasmith:
> I'm not reviewing the details, but the overall change looks great

I gave Claude similar feedback.

PR #118, opened 2025-04-22
feat: Cython Naming and File Structure Convention

rjwalters:
<h3>Summary</h3>
<p>This PR introduces a standardized naming and file structure convention for our Cython bindings to native C++ code. The aim is to make symbol origins obvious, reduce ambiguity between layers, and support maintainability and clarity throughout the codebase.</p>
<h3>File Structure</h3>
<p>We adopt a layered structure for bridging C++ and Python:</p>
<pre><code>cpp_code.hpp / cpp_code.cpp   # Native C++ declarations and implementations
cpp_code.pxd                  # Cython declarations for C++ types/functions (cimportable)
code.pxd                      # Internal Cython declarations to share across modules
code.pyx                      # Python-facing wrapper logic
</code></pre>
<h3>Naming Convention</h3>
<p>To distinguish clearly between native C++ symbols, Cython wrappers, and Python-facing interfaces, we apply consistent <strong>prefix-based naming</strong>:</p>

Layer | Prefix | Example Name | Notes
-- | -- | -- | --
C++ layer | cpp_ | cpp_Var, cpp_func() | Native C++ functions, variables, and helpers
C++ class types | Cpp | CppClass, CppVector | C++ classes/types exposed via cpp_code.pxd
Python wrapper | (clean) | Var, func() | Python-friendly interface exposed via .pyx
Internal Cython | (optional _) | _internal_func() | For internal use only, not part of public API

<h3>Rationale</h3>
<ul>
<li>
<p><strong>Clarity across boundaries</strong>: Prefixes make it immediately obvious whether you're interacting with native code or Cython/Python logic.</p>
</li>
<li>
<p><strong>Name conflict avoidance</strong>: Prevents collisions between C++ and Python-layer names.</p>
</li>
<li>
<p><strong>Tooling compatibility</strong>: Improves support for IDE symbol resolution and refactoring.</p>
</li>
<li>
<p><strong>Readability</strong>: Easier for new contributors to trace data flow and debug behavior across language boundaries.</p>
</li>
</ul>
<h3>Example</h3>
<p>Given a native C++ function:</p>
<pre><code class="language-cpp">// cpp_code.hpp
int cpp_add(int a, int b);
</code></pre>
<p>The corresponding Cython declaration and wrapper might look like:</p>
<pre><code class="language-cython">
# cpp_code.pxd
cdef extern from "cpp_code.hpp":
    int cpp_add(int a, int b)
</code><code class="language-cython">
# code.pyx
def add(a: int, b: int) -&gt; int:
    return cpp_add(a, b)
</code>
</pre>

<p>Now users of the Python API see a clean interface:</p>
<pre><code class="language-python">from mymodule.code import add
add(2, 3)  # → 5
</code></pre>


mettagrid/action_handler.pyx
sasmith, line 1: Generally we want to move business logic out of cython, so converting from .hpp to .pyx seems like a step in the wrong direction.

mettagrid/actions/attack.hpp
```diff
@@ -2,29 +2,28 @@
...
 #include "objects/agent.hpp"
 #include "objects/constants.hpp"
 #include "objects/metta_object.hpp"
 
-class Attack : public ActionHandler {
+class Attack : public CppActionHandler {
 public:
-    Attack(const ActionConfig& cfg, const std::string& action_name = "attack")
-        : ActionHandler(cfg, action_name) {
+    Attack(const cpp_ActionConfig& cfg, const std::string& action_name = "attack") : CppActionHandler(cfg, action_name)
```
sasmith, line 14: I'm not a fan of the `cpp_` naming in cpp files, since I haven't had experiences within cpp where I've gotten confused, and so the verbosity isn't something I find helpful. We're not writing any code where Python object make it to this depth (at least, if we are, I'm missing the fact). Where I'd be more actively supportive is in the .pyx layer, where we have both Python and cpp objects intermixed.

Aka, the rationale you have in the commit message makes sense in a general way, but I've only found it to apply in the pyx layer. If you've found it to be otherwise, I'd be interested to see examples.

rjwalters:
## 🚧 Re-evaluating C++ Testing Strategy

After working to integrate unit tests and benchmarks for our C++ code via Cython, I've hit several significant roadblocks that make this approach more effort than it's worth:

### Strict namespacing and exposure issues
Cython requires extensive ceremony to expose C++ methods cleanly. Even with care, method signatures, cdef/cpdef boundaries, and struct wrapping result in a lot of boilerplate and fragile interface code.

### Benchmarking limitations
pytest-benchmark cannot directly time cdef methods, and benchmarking C++ performance through Python wrappers forces a cythonize precompile step. This breaks the clean test/benchmark flow and undermines timing accuracy.

Given these limitations, I plan to drop this Cython-based effort and restart later using a more native and modular approach:

-  C++ native unit tests using gtest or Catch2
-  C++ native benchmarking using google/benchmark
-  A minimal and optional pybind11 interface to expose a lightweight Python-facing API for interop when needed

This change will:

-  Decouple testing from Python entirely
-  Make it easier to write and maintain idiomatic, fast C++ code
-  Allow us to reuse tooling used widely in the C++ ecosystem
-  Avoid the complexity of cythonize, conftest hooks, and Python import semantics

rjwalters:
moved to https://github.com/Metta-AI/metta/pull/197 and https://github.com/Metta-AI/metta/pull/198

